PR #32, opened 2022-03-14
removed step 2, similar to asp.net sample

gabriel-okteto:


pchico83:
@gabriel-okteto is this still valid? should we close it?

gabriel-okteto:
> @gabriel-okteto is this still valid? should we close it?

this should be closable

PR #62, opened 2022-04-08
Update manifest docs

AgustinRamiroDiaz:
Signed-off-by: Agustin Ramiro Diaz <agustin.ramiro.diaz@gmail.com>

### DIFF
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -124,7 +124,7 @@ It's usually a combination of `helm`, `kubectl`, and `okteto` commands.
 
 If you are planning to develop with your `docker-compose` only, you should [check out our "Getting Started" guide](/docs/using-dev-envs/)!
 
-If you created an `okteto` manifest to get the most out of okteto (like customizing the `dev` section) you can use the following syntax to deploy your `docker-compose`:
+You can also create an `okteto.yml` manifest to get even more out of okteto (e.g. to customize the folders that get synchronized, add extra environment variables, etc...). If you do that, you'll need to update the `deploy` section as follows:
 
 ```yaml
 deploy:


src/content/reference/manifest.mdx
```diff
@@ -144,6 +144,8 @@ deploy:
 #### Advanced deploy:
 
 There is an extended notation to deploy Docker Compose files and [endpoints](/docs/reference/compose/#endpoints-object-optional) as part of your development environment:
```
RinkiyaKeDad, line 146:
```suggestion
There is an extended notation to deploy Docker compose files and [endpoints](/docs/reference/compose/#endpoints-object-optional) as part of your development environment:
```

src/content/reference/manifest.mdx
```diff
@@ -144,6 +144,8 @@ deploy:
...
+
+The `services` field allows us to deploy only a subset of the services in the compose file. Only the `volumes` and `endpoints` of the services specified in the `services` field are deployed.
```
RinkiyaKeDad, line 148:
```suggestion
The `services` field allows us to deploy only a subset of the services in the Docker compose file. Only the `volumes` and `endpoints` of the services specified in the `services` field are deployed.  
```

src/content/reference/manifest.mdx
```diff
@@ -120,6 +120,14 @@ A list of commands to deploy your development environment.
 
 It's usually a combination of `helm`, `kubectl`, and `okteto` commands.
 
+#### Deploy a `docker-compose` file:
+
+```yaml
+deploy:
```
rberrelleza, line 126: I find thus very confusing, since we say in other places that all you need is the compose file. What's the benefit of having this in the deploy vs not having an okteto.yml file?

pchico83: You might want to define dev containers, add `dependencies` or deploy k8s resourcces

AgustinRamiroDiaz: It's part of what we accept in the schema. We could remove this to encourage users to use only a `docker-compose` or go directly to the advanced features (multiple composes, dependencies, k8s resources, only deploy subset...). But there'll be times when you want to create an `okteto manifest` to specify another section (like `dev`) and that's when this notation is useful

rberrelleza: I agree is useful, but we should tell our users why is useful, and make it clear that this is not required. 

I think that the use case of "customize the dev section" makes a lot of sense. The "deploy dependencies" or "k8s resources" sounds like a more advanced use case to me. I think that should be documented somewhere else.

AgustinRamiroDiaz: @rberrelleza I changed it to redirect users with only a docker-compose to its corresponding guide.

If someone is going through the `manifest schema reference`, I think they should see all the available options.
If not, I think we could split it into 2:
- a specification of the manifest schema
- a simpler user guide

src/content/reference/manifest.mdx
```diff
@@ -120,6 +120,18 @@ A list of commands to deploy your development environment.
 
 It's usually a combination of `helm`, `kubectl`, and `okteto` commands.
 
+#### Deploy a `docker-compose` file:
+
+If you are planning to develop with your `docker-compose` only, you should [check out our guide](/docs/tutorials/compose-getting-started/)!
+
+If you created an `okteto` manifest to get the most out of okteto (like customizing the `dev` section) you can use the following syntax to deploy your `docker-compose`:
```
rberrelleza, line 127:
```suggestion
You can also create an `okteto.yml` manifest to get even more out of okteto (e.g. to customize the folders that get synchronized, add extra environment variables, etc...). If you do that, you'll need to update the `deploy` section as follows:
```

rberrelleza (approved):

src/content/reference/manifest.mdx
```diff
@@ -120,6 +120,18 @@ A list of commands to deploy your development environment.
 
 It's usually a combination of `helm`, `kubectl`, and `okteto` commands.
 
+#### Deploy a `docker-compose` file:
+
+If you are planning to develop with your `docker-compose` only, you should [check out our guide](/docs/tutorials/compose-getting-started/)!
```
pchico83, line 125: can we link to the getting started guide? there is one option for compose there

PR #85, opened 2022-05-04
Update manifest for envvars with `-`

pchico83:


src/content/reference/manifest.mdx
```diff
@@ -70,6 +70,8 @@ In order to refer to these images from anywhere in your [Okteto Manifest](/docs/
 
 You can build all these images by running `okteto build`, or `okteto build xxx` to build a single one.
 
+> if a `<service>` name includes the symbol `-`, the corresponding `OKTETO_BUILD_<service>` environment variables substitude the symbol `-` by `_`.
```
rberrelleza, line 73: Can you add an example?

rberrelleza: It's not something that's easy to understand just from the text. If it doesn't fit as a note we can make it a subsection or something like it.

rberrelleza, line 73: maybe like this?
```suggestion
> if a `<service>` name includes the symbol `-`, the corresponding `OKTETO_BUILD_<service>` environment variables substitude the symbol `-` by `_`.  For example, f the service name is service-with-hyphen, the environment variable will be named $OKTETO_BUILD_SERVICE_WITH_HYPHEN.
```

rberrelleza:
@pchico83 can we merge this?

PR #90, opened 2022-05-09
point to community instead of cloud-feedback

rberrelleza:


pchico83 (approved):
@rberrelleza after merge, I am going to archive cloud-feedback, and ask on every issue to use community.okteto.com

PR #93, opened 2022-05-10
Added docs for config flags for volumeSnapshots

teresaromero:
Signed-off-by: Teresa Romero <teresa@okteto.com>

Added support for the new flags regarding VolumeSnapshots usage when creating PVC

### DIFF
--- a/src/content/enterprise/administration/configuration.mdx
+++ b/src/content/enterprise/administration/configuration.mdx
@@ -825,8 +825,8 @@ This feature requires having a CSI driver installed in your cluster.
 - `driver`: The name of the CSI driver used when creating snapshots.
 - `class`: The VolumeSnapshotClass of the volume snapshot.
 - `storageClass`: The storage class required by volumes initialized from snapshots (optional).
-- `enableNamespaceAccessValidation`: Enable check of namespace to allow only members from the namespace to use the volume snapshot. Default false.
-- `allowIDAnnotation`: Allow use of snapshot-id as source. Default true.
+- `enableNamespaceAccessValidation`: When enabled, only users that have access to the namespace where the volume snapshot is stored will be able to use it in their development environments. Default false.
+- `allowIDAnnotation`: Allow using your cloud provider's snapshot ID as the source of the data. Default true.
 
 ```yaml
 volumeSnapshots:
--- a/src/content/enterprise/administration/configuration.mdx
+++ b/src/content/enterprise/administration/configuration.mdx
@@ -825,6 +825,8 @@ This feature requires having a CSI driver installed in your cluster.
 - `driver`: The name of the CSI driver used when creating snapshots.
 - `class`: The VolumeSnapshotClass of the volume snapshot.
 - `storageClass`: The storage class required by volumes initialized from snapshots (optional).
+- `enableNamespaceAccessValidation`: Enable check of namespace to allow only members from the namespace to use the volume snapshot. Default false.
+- `allowIDAnnotation`: Allow use of snapshot-id as source. Default true.
 
 ```yaml
 volumeSnapshots:
--- a/src/content/enterprise/administration/volume-snapshots.mdx
+++ b/src/content/enterprise/administration/volume-snapshots.mdx
@@ -238,6 +238,7 @@ spec:
 ```
 
 If the annotation `dev.okteto.com/from-snapshot-namespace` is not defined, Okteto will default to the namespace of the new persistent volume claim.
+Access to snapshots from all namespaces of the cluster is enabled by default, if you want to limit this so users can only access snapshots from namespaces which they members, you can enable the setting `enableNamespaceAccessValidation: true` at the configuration `volumeSnapshots`.
 
 ## Using Storage Snapshots in your Development Environment
 
@@ -310,3 +311,4 @@ spec:
 </Tabs>
 
 As part of the creation of the persistent volume claim, Okteto will import the specified snapshot into your cluster using a `VolumeSnapshotContent` and will set the source of your persistent volume claim to this [VolumeSnapshotContent](https://kubernetes.io/docs/concepts/storage/volume-snapshots/), ready to be used by your development environment.
+Allowing volume snapshots from snapshot id is enabled by default, you can disable this feature by changing the config `allowIDAnnotation: false` at the cluster configuration `volumeSnappshots`.
--- a/src/content/enterprise/administration/volume-snapshots.mdx
+++ b/src/content/enterprise/administration/volume-snapshots.mdx
@@ -312,4 +312,4 @@ spec:
 
 As part of the creation of the persistent volume claim, Okteto will import the specified snapshot into your cluster using a `VolumeSnapshotContent` and will set the source of your persistent volume claim to this [VolumeSnapshotContent](https://kubernetes.io/docs/concepts/storage/volume-snapshots/), ready to be used by your development environment.
 
-Allowing volume snapshots from snapshot id is enabled by default, you can disable this feature by changing the config `allowIDAnnotation: false` at the cluster configuration `volumeSnappshots`.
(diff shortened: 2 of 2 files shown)


src/content/enterprise/administration/volume-snapshots.mdx
```diff
@@ -309,4 +310,6 @@ spec:
 </TabItem>
 </Tabs>
 
-As part of the creation of the persistent volume claim, Okteto will import the specified snapshot into your cluster using a `VolumeSnapshotContent` and will set the source of your persistent volume claim to this [VolumeSnapshotContent](https://kubernetes.io/docs/concepts/storage/volume-snapshots/), ready to be used by your development environment.
\ No newline at end of file
+As part of the creation of the persistent volume claim, Okteto will import the specified snapshot into your cluster using a `VolumeSnapshotContent` and will set the source of your persistent volume claim to this [VolumeSnapshotContent](https://kubernetes.io/docs/concepts/storage/volume-snapshots/), ready to be used by your development environment.
+
+Allowing volume snapshots from snapshot id is enabled by default, you can disable this feature by changing the config `allowIDAnnotation: false` at the cluster configuration `volumeSnappshots`.
```
rberrelleza, line 315: Why would someone want to disable this? Can we specify the scenario this enables?

maroshii, line 315: `volumeSnappshots` -> `volumeSnapshots`

teresaromero: Sure, added. The case where OE customer wants to prevent the usage of snapshots-id as source for PVCs

src/content/enterprise/administration/configuration.mdx
```diff
@@ -825,6 +825,8 @@ This feature requires having a CSI driver installed in your cluster.
 - `driver`: The name of the CSI driver used when creating snapshots.
 - `class`: The VolumeSnapshotClass of the volume snapshot.
 - `storageClass`: The storage class required by volumes initialized from snapshots (optional).
+- `enableNamespaceAccessValidation`: Enable check of namespace to allow only members from the namespace to use the volume snapshot. Default false.
```
rberrelleza, line 828:
```suggestion
- `enableNamespaceAccessValidation`: When enabled, only users that have access to the namespace where the volume snapshot is stored will be able to use it in their development environments. Default false.
```

src/content/enterprise/administration/configuration.mdx
```diff
@@ -825,6 +825,8 @@ This feature requires having a CSI driver installed in your cluster.
...
+- `allowIDAnnotation`: Allow use of snapshot-id as source. Default true.
```
rberrelleza, line 829: Is this called snapshot ID in all cloud providers?
```suggestion
- `allowIDAnnotation`: Allow using your cloud provider's snapshot ID as the source of the data. Default true.
```

src/content/enterprise/administration/volume-snapshots.mdx
```diff
@@ -238,6 +238,7 @@ spec:
 ```
 
 If the annotation `dev.okteto.com/from-snapshot-namespace` is not defined, Okteto will default to the namespace of the new persistent volume claim.
+Access to snapshots from all namespaces of the cluster is enabled by default, if you want to limit this so users can only access snapshots from namespaces which they members, you can enable the setting `enableNamespaceAccessValidation: true` at the configuration `volumeSnapshots`.
```
rberrelleza, line 241:
```suggestion
By default, all users in your cluster will be able to use any of the available snapshots as a data source for their development environments.  You can limit this via the `enableNamespaceAccessValidation: true` key in the  `volumeSnapshots` section of the configuration.
```

src/content/enterprise/administration/volume-snapshots.mdx
```diff
@@ -309,4 +310,6 @@ spec:
 </TabItem>
 </Tabs>
 
-As part of the creation of the persistent volume claim, Okteto will import the specified snapshot into your cluster using a `VolumeSnapshotContent` and will set the source of your persistent volume claim to this [VolumeSnapshotContent](https://kubernetes.io/docs/concepts/storage/volume-snapshots/), ready to be used by your development environment.
\ No newline at end of file
+As part of the creation of the persistent volume claim, Okteto will import the specified snapshot into your cluster using a `VolumeSnapshotContent` and will set the source of your persistent volume claim to this [VolumeSnapshotContent](https://kubernetes.io/docs/concepts/storage/volume-snapshots/), ready to be used by your development environment.
+
+Allowing volume snapshots from a snapshot-id is enabled by default, you can disable this feature by changing the configuration `allowIDAnnotation` to false in case you want to prevent PVCs created from any snapshot-id at the cluster.
```
rberrelleza, line 315:
```suggestion
By default, anyone can use any of your cloud provider's snapshot ID as a data source. You can disable this by changing the configuration `allowIDAnnotation ` to `false` in the `volumeSnapshots` section of the configuration.
```

src/content/enterprise/administration/configuration.mdx
rberrelleza: @teresaromero when this is set to false what happens? users can only use PVCs as the source?

src/content/enterprise/administration/volume-snapshots.mdx
rberrelleza: is `data source` the right term?

src/content/enterprise/administration/configuration.mdx
teresaromero: they can use snapshots with the namespace and name annotation, or the native k8s dataSource if the snapshot belongs to the same namespace 

```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    # OPTION 1 retrieve snapshot from other namespace
    dev.okteto.com/from-snapshot-name: <name>
    dev.okteto.com/from-snapshot-namespace: <namespace>
    # OPTION 2 retrieve snapshot from id
    dev.okteto.com/from-snapshot-id: <id>
  name: pvc-from-snapshot
spec:
  storageClassName: csi-okteto-standard
  # OPTION 3 retrieve snapshot k8s native
  # snapshot has to be at the same namespace as the PVC is created
  dataSource:
    name: pvc-test-snapshot
    kind: VolumeSnapshot
    apiGroup: snapshot.storage.k8s.io
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
```

src/content/enterprise/administration/volume-snapshots.mdx
teresaromero: i guess it is, as when using the native k8s key to provide this at the pvc spec is dataSource

src/content/enterprise/administration/volume-snapshots.mdx
```diff
@@ -312,4 +312,4 @@ spec:
 
 As part of the creation of the persistent volume claim, Okteto will import the specified snapshot into your cluster using a `VolumeSnapshotContent` and will set the source of your persistent volume claim to this [VolumeSnapshotContent](https://kubernetes.io/docs/concepts/storage/volume-snapshots/), ready to be used by your development environment.
 
-Allowing volume snapshots from a snapshot-id is enabled by default, you can disable this feature by changing the configuration `allowIDAnnotation` to false in case you want to prevent PVCs created from any snapshot-id at the cluster.
\ No newline at end of file
+By default, anyone can use any of your cloud provider's snapshot ID as a data source. You can disable this by changing the configuration `allowIDAnnotation ` to `false` in the `volumeSnapshots` section of the configuration.
```
rberrelleza, line 315: @teresaromero when this feature is disabled, what option do the user has then? only use the dev.okteto.com/from-snapshot-namespace annotation?

jmacelroy (approved):

Issue #94, opened 2022-05-10
Rename Gitlab > GitLab


davidpthomas:
Typo fix across documentation.  Rename "Gitlab" > "GitLab"

PR #95, opened 2022-05-10
Renaming Gitlab > GitLab in all docs

davidpthomas:
Fixing a typo in doc repo.

dt-work:okteto-docs dt$ for i in $(find . -type f | xargs grep "Gitlab" |  awk -F: '{print $1}' | uniq); do sed -e 's/Gitlab/GitLab/g' $i > $i.sed && mv $i.sed $i;  done;

Signed-off-by: David 'DT' Thomas <dt@okteto.com>


davidpthomas:
related to #94

rberrelleza (approved):

PR #98, opened 2022-05-10
update docs header to remove options

rberrelleza:
I like the aesthetic of Gitlab docs. I think is clean and less distracting.
image

@huguestennier would it be possible to copy their layout, and have the search on the left (so it's more obvious). And a more noticeable button for the free trial link?

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -56,7 +56,7 @@ module.exports = {
           target: '_self'
         },
         {
-          href: 'https://okteto.com/community/',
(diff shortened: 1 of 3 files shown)


huguestennier:
@rberrelleza Done!

We can also move search as the first item on the left if we move it up in the config file. What do you think?

Screen Shot 2022-05-10 at 9 01 49 PM

rberrelleza:
@huguestennier yeah, good idea, let's do it.

PR #102, opened 2022-06-01
docs: add note regarding the use of CNAME records

jpf-okteto:
This PR adds a note discouraging the use of `CNAME` records for `*` that point to the public hostname of AWS Elastic Load Balancers.

Okteto Enterprise automatic certificate management uses cert-manager to issue the wildcard certificate used by the deployment. This certificate is issued by Let's Encrypt as default and uses ACME dns01 verification challenge, which consists of creating a `TXT` dns record for the Okteto domain.

If the user creates a `CNAME` dns record for `*`, it will shadow any other dns records, including `TXT`. This blocks the issue/renewal process as the verification can no longer be achieved.

Reasons why a user may set a `CNAME` record include the scenario where they manage DNS from a different AWS account, making it impossible to set an `A` dns record with an alias, or where they manage DNS with a different service/software rather than AWS Route53.

The best way to solve the issue for these scenarios would be to use AWS Network Load Balancers, which expose a public static IP.

### DIFF
--- a/src/content/enterprise/eks.mdx
+++ b/src/content/enterprise/eks.mdx
@@ -257,6 +257,8 @@ okteto-ingress-nginx-controller  LoadBalancer   10.0.7.73    a519c8b3b27f94...el
 
 You'll need to take the `EXTERNAL-IP` address, and add it to your DNS for the domain you have chosen to use. In Route 53, this is done by creating an `A` record with the name  `*`, pointing to the alias of the Elastic Load Balancer.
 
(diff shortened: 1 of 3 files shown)


RinkiyaKeDad (commented):
We're no longer updating the older doc versions from what I remember, right @rberrelleza?

rberrelleza:
Correct, but I think it's ok since this is an issue with AWS and not with our product.

rberrelleza (approved):

PR #107, opened 2022-06-07
remove "Okteto Cloud" page from docs

rberrelleza:
As we move towards a model where the docs are the same for Okteto Cloud, Okteto for Teams or Enteprise (except for install instructions), I want to start removing any pages that are specific to "Okteto Cloud" or "Okteto Enterprise" if possible, and start talking about "Okteto" and perhaps, "Okteto Self-Hosted".  This is an example of this. 

Are we ok with removing this? Another option is to update https://www.okteto.com/docs/welcome/overview/ to include the list of features that today lives in https://www.okteto.com/docs/cloud/ (since all of those features are applicable for both managed or self-hosted)


RinkiyaKeDad:
I like it! It makes sense and will be much easier for users to understand. 

> Another option is to update https://www.okteto.com/docs/welcome/overview/ to include the list of features that today lives in https://www.okteto.com/docs/cloud/ (since all of those features are applicable for both managed or self-hosted)

I picture the welcome page as a very high-level introduction to our product, so I wouldn't recommend adding features or any other technical stuff there. WDYT?

rberrelleza:
> I like it! It makes sense and will be much easier for users to understand.
> 
> > Another option is to update https://www.okteto.com/docs/welcome/overview/ to include the list of features that today lives in https://www.okteto.com/docs/cloud/ (since all of those features are applicable for both managed or self-hosted)
> 
> I picture the welcome page as a very high-level introduction to our product, so I wouldn't recommend adding features or any other technical stuff there. WDYT?

sounds good

RinkiyaKeDad:
@rberrelleza there are merge conflicts in this one, plus https://github.com/okteto/docs/pull/118 covers the changes you've suggested here. Are we okay with closing this?

rberrelleza:
> @rberrelleza there are merge conflicts in this one, plus #118 covers the changes you've suggested here. Are we okay with closing this?

I'll fix the merge conflicts. Let's do the change on smaller increments, it's easier to review and ensure that there aren't any broken links.

src/content/cloud.mdx
```diff
@@ -1,28 +0,0 @@
...
-title: Okteto Cloud
-description: Okteto Cloud gives you free access to secure Kubernetes namespaces, fully integrated with remote development capabilities
-sidebar_label: Okteto Cloud
-id: cloud
----
-
-[Okteto Cloud](https://cloud.okteto.com/#/?origin=docs) gives you free access to secure Kubernetes namespaces, fully integrated with remote development capabilities.
-Develop your applications in Okteto Cloud and forget about slow and tedious local development forever.
-
-With Okteto Cloud you can:
```
RinkiyaKeDad, line 11: My only concern is not having this list of features anywhere. Do you think they're essential? 

If yes, maybe we can have them on the Dev Environments Overview page since most of them are relevant for that?

RinkiyaKeDad (approved):

src/content/cloud.mdx
rberrelleza: I don't think they are that relevant, since the features are described on the docs. But maybe a quick summary?

PR #112, opened 2022-06-07
feat: add global forward doc

AdrianPedriza:
Signed-off-by: adripedriza <adripedriza@gmail.com>

Added documentation related to the global forward. Merge docs when added feat to a new CLI release.

### DIFF
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -634,8 +634,6 @@ forward:
       app: db
 ```
 
-When an `okteto up` session is already making use of the ports declared in the global forward section and another `okteto up` session is launched, the latter will see that global ports are already allocated by the previous session so it will keep retrying to reopen the forward (in case the initial session terminates and deallocates the ports).
-
 ### icon (string, optional)
 
 The icon associated to your development environmen in the Okteto Dashboard (optional).
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -614,7 +614,7 @@ Sets the working directory of your development container.
 
 Port forwards activated on any `okteto up` command, independently of the dev container name.
 
-As in [forward](#forward-string-optional) in dev section, several formats are accepted for this, but it is necessary to indicate the name of the service to be forwarded using the notation `localPort:remoteService:remotePort`.
+As in [forward](#forward-string-optional) in dev section, several formats are accepted for this, but it is necessary to indicate the name of the service to be forwarded using the `localPort:remoteService:remotePort` notation.
 
 ```yaml
 forward:
@@ -636,7 +636,7 @@ forward:
 
 Using a global forward, Okteto handles port collision if two `okteto up` sessions are running at the same time. 
 
-When a `okteto up` session is already making use of those ports declared in the global forward section and another `okteto up` session runs, the latter will see that global ports are already allocated by the previous session so it will retry every so often to reopen the forward (in case the initial session terminates and de-allocate ports).
+When an `okteto up` session is already making use of the ports declared in the global forward section and another `okteto up` session is launched, the latter will see that global ports are already allocated by the previous session so it will keep retrying to reopen the forward (in case the initial session terminates and deallocates the ports).
 
 ### icon (string, optional)
 
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -612,29 +612,29 @@ Sets the working directory of your development container.
 
 ### forward ([string], optional)
 
-Port forwards activated on any `okteto up` command, independently of the dev container name.
+Using a global forward, Okteto handles port collision if two `okteto up` sessions are running at the same time. 
 
-As in [forward](#forward-string-optional) in dev section, several formats are accepted for this, but it is necessary to indicate the name of the service to be forwarded using the `localPort:remoteService:remotePort` notation.
+As in [forward](#forward-string-optional) in dev section, several formats are accepted for this section.
 
 ```yaml
 forward:
   - 5432:postgres:5432
+  - localPort: 8080
+    remotePort: 80
+    name: app
 ```
+
+As we can see in the examples, in this case we need to indicate the name of the service to be forwarded.
 
 You can also indicate the service to expose using a label selector.
 
 ```yaml
 forward:
-  - localPort: 8080
-    remotePort: 80
-    name: app
   - localPort: 5432
     remotePort: 5432
     labels:
       app: db
 ```
-
-Using a global forward, Okteto handles port collision if two `okteto up` sessions are running at the same time. 
 
 When an `okteto up` session is already making use of the ports declared in the global forward section and another `okteto up` session is launched, the latter will see that global ports are already allocated by the previous session so it will keep retrying to reopen the forward (in case the initial session terminates and deallocates the ports).
 
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -614,7 +614,24 @@ Sets the working directory of your development container.
 
 You can also declare global forwards that will be used by all dev environments.
 
-As in `forward` in dev section, several formats are accepted for this, but it is necessary to indicate the name of the service to be forwarded.
+As in `forward` in dev section, several formats are accepted for this, but it is necessary to indicate the name of the service to be forwarded using the notation `localPort:remoteService:remotePort`.
+
+```yaml
+forward:
+  - 5432:postgres:5432
+  - localPort: 27017
+    remotePort: 27017
+    name: mongodb
+```
+
+The global forwards can be used by each declared development environment and can be used at the same time without any port collisions. 
+
+> Common uses of port forwarding are:
+>
+> 1.  Access a service via `localhost` instead of via an ingress
+> 1.  Reuse port forward for more than one dev environment
+
+If a process is already using the ports you want to use for global forwarding, it will be performed once they are released.
 
 ### icon (string, optional)
 
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -636,7 +636,7 @@ forward:
 
 Using a global forward, Okteto handles port collision if two `okteto up` sessions are running at the same time. 
 
-If a `okteto up` session is already making use of those ports declared in the global forward section and another `okteto up` session runs, the latter will see that global ports are already allocated by the previous session so it will retry every so often to reopen the forward (in case the initial session terminates and de-allocate ports).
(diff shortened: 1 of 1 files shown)


src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,12 @@ volumes:
 
 Sets the working directory of your development container.
 
+### forward ([string], optional)
+
+You can also declare global forwards that will be used by all dev environments.
+
+As in `forward` in dev section, several formats are accepted for this, but it is necessary to indicate the name of the service to be forwarded.
```
pchico83, line 617: Can we document the different formats and add samples here?

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,29 @@ volumes:
...
 Sets the working directory of your development container.
 
+### forward ([string], optional)
+
+You can also declare global forwards that will be used by all dev environments.
+
+As in `forward` in dev section, several formats are accepted for this, but it is necessary to indicate the name of the service to be forwarded using the notation `localPort:remoteService:remotePort`.
+
+```yaml
+forward:
```
pchico83, line 620: Let's document the different notations in different samples, like here:
https://www.okteto.com/docs/reference/manifest/#forward-string-optional

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,29 @@ volumes:
 
 Sets the working directory of your development container.
 
+### forward ([string], optional)
+
+You can also declare global forwards that will be used by all dev environments.
+
+As in `forward` in dev section, several formats are accepted for this, but it is necessary to indicate the name of the service to be forwarded using the notation `localPort:remoteService:remotePort`.
```
pchico83, line 617: Can link to the `dev.forward` section here

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,29 @@ volumes:
 
 Sets the working directory of your development container.
 
+### forward ([string], optional)
+
+You can also declare global forwards that will be used by all dev environments.
```
pchico83, line 615:
```suggestion
Port forwards activated on any `okteto up` command, independently of the dev container name.
```

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,29 @@ volumes:
...
+
+```yaml
+forward:
+  - 5432:postgres:5432
+  - localPort: 27017
+    remotePort: 27017
+    name: mongodb
+```
+
+The global forwards can be used by each declared development environment and can be used at the same time without any port collisions.
```
pchico83, line 627: I would skip the rest and just say that okteto handes these forwards to avoid port collisions if two `okteto up` sessions are running at the same time

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,32 @@ volumes:
...
+  - localPort: 8080
+    remotePort: 80
+    name: app
+  - localPort: 5432
+    remotePort: 5432
+    labels:
+      app: db
+```
+
+Using a global forward, Okteto handles port collision if two `okteto up` sessions are running at the same time.
```
rberrelleza, line 637: how does Okteto handle the collision?

AdrianPedriza: if a `okteto up` session is already making use of those ports declared in the global forward section and another `okteto up` session runs, the latter will see that they are already allocated (but will be able to make use of them because the forward has already been opened by the previous session) so it will retry (loop) every so often to reopen the forward (in case the initial session terminates).

rberrelleza: can we include this explanation int he docs? Just saying "handles port collision" is not very useful for our users. We should mention which port takes precedence (I think is the first), and how the ports of the second okteto up will be ignored until the first one is stopped.

pchico83:
@rberrelleza can we merge this one?

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,34 @@ volumes:
...
+    name: app
+  - localPort: 5432
+    remotePort: 5432
+    labels:
+      app: db
+```
+
+Using a global forward, Okteto handles port collision if two `okteto up` sessions are running at the same time. 
+
+When a `okteto up` session is already making use of those ports declared in the global forward section and another `okteto up` session runs, the latter will see that global ports are already allocated by the previous session so it will retry every so often to reopen the forward (in case the initial session terminates and de-allocate ports).
```
RinkiyaKeDad, line 639:
```suggestion
When an `okteto up` session is already making use of the ports declared in the global forward section and another `okteto up` session is launched, the latter will see that global ports are already allocated by the previous session so it will keep retrying to reopen the forward (in case the initial session terminates and deallocates the ports).
```

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,34 @@ volumes:
 
 Sets the working directory of your development container.
 
+### forward ([string], optional)
+
+Port forwards activated on any `okteto up` command, independently of the dev container name.
+
+As in [forward](#forward-string-optional) in dev section, several formats are accepted for this, but it is necessary to indicate the name of the service to be forwarded using the notation `localPort:remoteService:remotePort`.
```
RinkiyaKeDad, line 617:
```suggestion
As in [forward](#forward-string-optional) in dev section, several formats are accepted for this, but it is necessary to indicate the name of the service to be forwarded using the `localPort:remoteService:remotePort` notation.
```

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,34 @@ volumes:
 
 Sets the working directory of your development container.
 
+### forward ([string], optional)
+
+Port forwards activated on any `okteto up` command, independently of the dev container name.
```
rberrelleza, line 615: What do we mean by this? I'm not sure what the "dev container name" refers to here.

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,34 @@ volumes:
...
+
+As in [forward](#forward-string-optional) in dev section, several formats are accepted for this, but it is necessary to indicate the name of the service to be forwarded using the `localPort:remoteService:remotePort` notation.
```
rberrelleza, line 617: I find this confusing. I think we can split this in two paragraphs: one for saying that this uses the same format of the other forward, and a second to talk about the service format?

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,34 @@ volumes:
...
+  - localPort: 8080
+    remotePort: 80
+    name: app
+  - localPort: 5432
+    remotePort: 5432
+    labels:
+      app: db
+```
+
+Using a global forward, Okteto handles port collision if two `okteto up` sessions are running at the same time.
```
rberrelleza, line 637: I think should be the first thing in the section, since it explains how this is different from the regular forward

src/content/reference/manifest.mdx
AdrianPedriza: I mean that no matter which service is selected for `okteto up <svc>`, if the manifest sets a global forward, it will be used for any `okteto up` using that manifest.

src/content/reference/manifest.mdx
AdrianPedriza: and remove sentence `Port forwards activated on any okteto up command, independently of the dev container name.`? I think they both express the same thing.

rberrelleza: yeah let's do that.

src/content/reference/manifest.mdx
rberrelleza: then let's say "global port forwards are always activated, independently of the service you are developing on". 

```suggestion
Global port forwards are always activated, independently of the service you are developing on.
```

rberrelleza (commented):
On which version of Okteto is this available? on 0.15?

AdrianPedriza:
> On which version of Okteto is this available? on 0.15?

I am not sure about it. It was added in okteto CLI 2.5.0

pchico83:
> > On which version of Okteto is this available? on 0.15?
> 
> I am not sure about it. It was added in okteto CLI 2.5.0

It's available from chart 0.15:
https://www.okteto.com/docs/enterprise/install/releases/#0150

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,34 @@ volumes:
 
 Sets the working directory of your development container.
 
+### forward ([string], optional)
+
+Using a global forward, Okteto handles port collision if two `okteto up` sessions are running at the same time.
```
rberrelleza, line 615:
```suggestion
When declaring a global forward, Okteto will automatically handle port collision when two or more `okteto up` sessions are running simultaneously. If the`okteto` detects that the port is already in use, and said port is defined as global forward,  `okteto up` will ignore the port collision and continue the `up` sequence. If the port is later available, `okteto` will automatically connect to it without interrupting the session.   This is disabled by default. 
```

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,34 @@ volumes:
 
 Sets the working directory of your development container.
 
+### forward ([string], optional)
+
+Using a global forward, Okteto handles port collision if two `okteto up` sessions are running at the same time. 
+
+As in [forward](#forward-string-optional) in dev section, several formats are accepted for this section.
```
rberrelleza, line 617:
```suggestion
Global forward uses the same format than [forward](#forward-string-optional):
```

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,34 @@ volumes:
...
+
+```yaml
+forward:
+  - 5432:postgres:5432
+  - localPort: 8080
+    remotePort: 80
+    name: app
+```
+
+As we can see in the examples, in this case we need to indicate the name of the service to be forwarded.
```
rberrelleza, line 627: Let's remove this line. 
```suggestion

```

src/content/reference/manifest.mdx
```diff
@@ -610,6 +610,34 @@ volumes:
...
+
+```yaml
+forward:
+  - localPort: 5432
+    remotePort: 5432
+    labels:
+      app: db
+```
+
+When an `okteto up` session is already making use of the ports declared in the global forward section and another `okteto up` session is launched, the latter will see that global ports are already allocated by the previous session so it will keep retrying to reopen the forward (in case the initial session terminates and deallocates the ports).
```
rberrelleza, line 639: I think we don't need this if we go with my suggestion above. 
```suggestion

```

src/content/reference/manifest.mdx
rberrelleza: @pchico83 @jmacelroy @AdrianPedriza what do you think about my explanation here? This one is tricky, i want to make sure is clear for everyone that reads it.

AdrianPedriza: much more accurate but I'm not quite clear what the last sentence (`This is disabled by default`) refers to.

pchico83 (approved):

PR #113, opened 2022-06-07
update sample to fix typo

rberrelleza:
@pchico83 can you confirm that this is the correct format? checkr reported this.

### DIFF
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -108,10 +108,8 @@ dependencies:
     manifest: okteto.yml
     branch: main
     variables:
-      - name: ENVIRONMENT
-        value: development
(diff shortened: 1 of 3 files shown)


pchico83:
> @pchico83 can you confirm that this is the correct format? checkr reported this.

@teresaromero @jLopezbarb could you review this? do we support any other format like this one?
```
    variables:
      - ENVIRONMENT=development
      - DEBUG=true
```

teresaromero:
variables at dependencies are supported as Environment in other parts of the manifest, we currently **not** support:

```
variables:
  - name: ENVIRONMENT
  - value: development
```

We do support:

```
variables:
  - ENVIRONMENT=development
```

```
variables:
  ENVIRONMENT: development
```

PR #120, opened 2022-06-26
Update docs to document that the Helm catalog is disabled by default

pchico83:
We are disabling the Helm catalog by default in self-hosted installations.
Our catalog requires to install custom CRDs that make the whole installation process more complicated

### DIFF
--- a/src/content/cloud/deploy-from-helm.mdx
+++ b/src/content/cloud/deploy-from-helm.mdx
@@ -10,7 +10,8 @@ import Image from '@theme/Image';
 You can launch additional services for your development environments on Okteto Cloud using Helm repositories.
 Okteto automates all the processes needed to deploy, configure, upgrade, and destroy your Helm releases, so you can focus on building amazing applications.
 
-You can also [list your own repository under the **Helm Catalog**](/docs/enterprise/administration/configuration/#applications) tab of the **Deployment** menu. This would give you the same one-click deployment experience for your own services when using [Okteto Self-Hosted](/docs/enterprise/).
+You can also list your own charts under the [Helm Catalog](/docs/enterprise/administration/configuration/#applications) tab of the **Launch Dev Environment** menu.
+This would give you the same one-click deployment experience for your own services when using [Okteto Self-Hosted](/docs/enterprise/).
 
 ## Deploy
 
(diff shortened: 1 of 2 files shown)


src/content/cloud/deploy-from-helm.mdx
```diff
@@ -10,6 +10,8 @@ import Image from '@theme/Image';
 You can launch additional services for your development environments on Okteto Cloud using Helm repositories.
 Okteto automates all the processes needed to deploy, configure, upgrade, and destroy your Helm releases, so you can focus on building amazing applications.
 
+You can also [bring your own Helm repository](/docs/enterprise/administration/configuration/#applications) to the catalog and get the same one-click deployment experience for your own services when using [Okteto Self-Hosted](/docs/enterprise/).
```
RinkiyaKeDad, line 13: It was a bit confusing to me what "catalog" meant until I looked at the image below in the deploy section, how about something like this:

```suggestion
You can also [list your own repository under the **Helm Catalog**](/docs/enterprise/administration/configuration/#applications) tab of the **Deployment** menu. This would give you the same one-click deployment experience for your own services when using [Okteto Self-Hosted](/docs/enterprise/).
```

RinkiyaKeDad (approved):

rberrelleza:
On which version is this going to be enabled?

pchico83:
> On which version is this going to be enabled?

@rberrelleza it will go with the next release

rberrelleza:
> > On which version is this going to be enabled?
> 
> @rberrelleza it will go with the next release

we need to create the new version of the docs before we merge these PRs. Is it possible to create a single branch for all the changes that are going with 0.14 instead of single PR per change?  We have a lot of changes coming into the docs (due to the pricing changes) so it would be easier if we only have one branch for 0.14 to worry about.

pchico83:
> > > On which version is this going to be enabled?
> > 
> > 
> > @rberrelleza it will go with the next release
> 
> we need to create the new version of the docs before we merge these PRs. Is it possible to create a single branch for all the changes that are going with 0.14 instead of single PR per change? We have a lot of changes coming into the docs (due to the pricing changes) so it would be easier if we only have one branch for 0.14 to worry about.

@rberrelleza I have updated the PR to send it against the `release-0.14` branch.
We can merge all chart 0.14 related PRs there, and merge `release-0.14` to `main` once the chart release is available

PR #130, opened 2022-07-03
Document "env" for different components

pchico83:
There are some scenarios when configuring custom envvars for different components is needed, for example, to configure outgoing requests via a proxy.


src/content/enterprise/administration/configuration.mdx
```diff
@@ -226,6 +226,7 @@ It is possible to use separate cloud providers for DNS than for storage if neede
 The API service. Account and Kubernetes credentials management, namespace creation, and sharing, deployment via the catalog, etc...
 
 - `annotations`: Annotations to add to the API pods.
+- `env`: Environment variables to add to the API pods.
```
rberrelleza, line 229:
```suggestion
- `env`: Environment variables to be set on the API containers.
```

src/content/enterprise/administration/configuration.mdx
```diff
@@ -287,11 +288,30 @@ autoscaler:
...
+- `installer.env`: Environment variables to add to the [installer jobs](/docs/enterprise/administration/configuration/#installer).
+- `installer.image`: Custom runtime image for the [installer jobs](/docs/enterprise/administration/configuration/#installer). Defaults to [okteto/pipeline-runner](https://github.com/okteto/pipeline-runner).
+- `installer.labels`: Labels to add to the [installer jobs](/docs/enterprise/administration/configuration/#installer).
+- `image.registry`: Image registry URL for the [api](/docs/enterprise/administration/configuration/#api) and [webhook](/docs/enterprise/administration/configuration/#webhook) services.
+- `image.tag`: Image tag for the [api](/docs/enterprise/administration/configuration/#api) and [webhook](/docs/enterprise/administration/configuration/#webhook) services.
+
+```yaml
+backend:
+  installer:
+    image: okteto/pipeline-runner:1.0.0
```
rberrelleza, line 306: can you also add an example for the env vars?

src/content/enterprise/administration/configuration.mdx
```diff
@@ -287,11 +288,30 @@ autoscaler:
...
+### backend
+
+Common configuration for backend services.
+
+- `pipelineInstaller.image`: Base image for the binaries needed by the [installer jobs](/docs/enterprise/administration/configuration/#installer).
+- `installer.annotations`: Annotations to add to the [installer jobs](/docs/enterprise/administration/configuration/#installer).
+- `installer.env`: Environment variables to add to the [installer jobs](/docs/enterprise/administration/configuration/#installer).
+- `installer.image`: Custom runtime image for the [installer jobs](/docs/enterprise/administration/configuration/#installer). Defaults to [okteto/pipeline-runner](https://github.com/okteto/pipeline-runner).
+- `installer.labels`: Labels to add to the [installer jobs](/docs/enterprise/administration/configuration/#installer).
+- `image.registry`: Image registry URL for the [api](/docs/enterprise/administration/configuration/#api) and [webhook](/docs/enterprise/administration/configuration/#webhook) services.
```
rberrelleza, line 300: I don't think we should document this, unless we expect this to be changed by the user right?

src/content/enterprise/administration/configuration.mdx
```diff
@@ -287,11 +288,30 @@ autoscaler:
   slackWebhook:
 ```
 
+### backend
+
+Common configuration for backend services.
+
+- `pipelineInstaller.image`: Base image for the binaries needed by the [installer jobs](/docs/enterprise/administration/configuration/#installer).
+- `installer.annotations`: Annotations to add to the [installer jobs](/docs/enterprise/administration/configuration/#installer).
+- `installer.env`: Environment variables to add to the [installer jobs](/docs/enterprise/administration/configuration/#installer).
```
rberrelleza, line 297: Why are we putting the `installer` configuration inside `backend`? I think it's better if we have a `installer` section the same way that we have an `api` section, and we avoid unnecessary hierarchies.

src/content/enterprise/administration/configuration.mdx
```diff
@@ -287,11 +288,30 @@ autoscaler:
...
+  installer:
+    image: okteto/pipeline-runner:1.0.0
+```
+
 ### buildkit
 
 The build service. It's used in combination with `okteto build` to build containers directly in the cluster.
 
 - `annotations`: Annotations to add to the buildkit pods.
+- `env`: Environment variables to add to the buildkit pods.
```
rberrelleza, line 314: This change suggestion applies to all the other ones as well, since technically the env var is applied to the container and not to the entire pod.
```suggestion
- `env`: Environment variables to be set on the buildkit containers.
```

src/content/enterprise/administration/configuration.mdx
rberrelleza, line 229: I think we should follow the pattern that https://artifacthub.io/packages/helm/bitnami/mysql or [gitlab](https://artifacthub.io/packages/helm/gitlab/gitlab?modal=values&path=global.extraEnv) uses and call these `extraEnvVars` or ` extraEnv` so it's more obvious that this will add to the existing (whatever this might be) instead of replacing it entirely (which is what I'd think when I see a property called `env`).

src/content/enterprise/administration/configuration.mdx
pchico83: which one? the `image`?
we do expect people to change the image, there are two use cases here:
- companies not pulling from dockerhub
- get access to a commit not released in a chart release yet

src/content/enterprise/administration/configuration.mdx
pchico83: Yes, but this is the way it's implemented in the chart from a long time ago. we should refactor that, but I would document it the way it is for now, thoughts?

src/content/enterprise/administration/configuration.mdx
pchico83: I agree. Let me see if we can change that before 0.14 is released

src/content/enterprise/administration/configuration.mdx
pchico83: I removed the `backend` section, we should deprecate it at some point. I moved `labels`, `annotations` and `extraEnv` to the installer section because they are new fields.
The `backend` section is to overwrite our images, we should move that to every component, but for now let's keep it undocumented

src/content/enterprise/administration/configuration.mdx
rberrelleza: It was implemented, but it was never documented. If we want to expose this to our customers,I think  it's "cheaper" to do it now rather than to deal with a migration/breaking change later.

src/content/enterprise/administration/configuration.mdx
rberrelleza: Sounds good!

PR #131, opened 2022-07-04
Remove dependency on "okteto init"

pchico83:
Our current getting started guides are kind of broken, because `okteto init` is not in a good shape yet.

This PR does the following:
- Unify the compose and helm samples by using `yarn start` as the `api` dev container command.
- Commit `okteto.yml` to the okteto/movies-with-helm repo
- BE more explicit about the need of configuring an okteto manifest


src/content/getting-started.mdx
```diff
@@ -66,21 +66,25 @@ $ okteto context list
...
+For Okteto CLI to know how to launch your Development Environment, it needs an [Okteto manifest](/docs/cloud/okteto-cli/#okteto-manifest). This manifest allows you to provide configuration for your Development Environments.
+Okteto CLI will analyze any [existing manifests](/docs/cloud/okteto-cli/#deployment-manifests) (Helm, Kubernetes, etc.) in your source code directory and help you write an Okteto manifest for your application.
+The Okteto manifest is necessary in all cases except if you already have a Docker compose file configured for your application.
 
-If Okteto CLI can't find the Okteto manifest for your application, the manifest generation step will happen automatically when we launch our Development Environment. You can also choose to explicitly generate the Okteto manifest using: 
+If Okteto CLI can't find the Okteto manifest for your application, `okteto init` helps you to create the first version of your Okteto manifest:
```
RinkiyaKeDad, line 78: Should we add a line here making it clear that users following this guide don't have to run `okteto init`?

```suggestion
If you don't have an Okteto manifest for your application, you can run `okteto init`. This generates the Okteto manifest, which you can then edit based on your needs. 

For the Getting Started guide, we already have the Okteto manifest configured for our application, so this step won't be needed.
```

src/content/getting-started.mdx
```diff
@@ -66,21 +66,25 @@ $ okteto context list
...
 
 ```console
 $ okteto init
 ```
```
RinkiyaKeDad, line 82: We can remove this now I guess.

RinkiyaKeDad (approved):

PR #141, opened 2022-07-13
Remove "kubectl" command to retrieve ssh key

pchico83:
Now we have the ability to download the SSH key from the Admin View, we shouldn't document how to access it using `kubectl`. This is too dependent on the implementation of the feature, and it's a poorer experience

### DIFF
--- a/src/content/administration/private-repositories/ssh-key.mdx
+++ b/src/content/administration/private-repositories/ssh-key.mdx
@@ -20,14 +20,6 @@ The public SSH is available on the `Settings` section on the `Admin -> General` 
 
 <p align="center"><Image src="/docs/administration/private-repositories/images/private-repository-ssh-key.png" alt="show SSH key" width="676" /></p>
 
-### Retrive the SSH key using kubectl <TiersList tiers="Self-Hosted" />
-
-You can also retreive the public SSH key by running the command below:
-
-```console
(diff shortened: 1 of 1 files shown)


src/content/administration/private-repositories/ssh-key.mdx
```diff
@@ -20,14 +20,6 @@ The public SSH is available on the `Settings` section on the `Admin -> General`
 
 <p align="center"><Image src="/docs/administration/private-repositories/images/private-repository-ssh-key.png" alt="show SSH key" width="676" /></p>
 
-### Retrive the SSH key using kubectl <TiersList tiers="Self-Hosted" />
```
rberrelleza, line 23: @pchico83 in which version is this available? if it's coming on 0.14, can you merge this on the 0.14 branch?

rberrelleza: And can you add how to do it via the UI?

rberrelleza (approved):

src/content/administration/private-repositories/ssh-key.mdx
pchico83: @rberrelleza this is available from many versions ago, the feature has at least 1 year. The UI flow is right above the section I am removing

PR #142, opened 2022-07-14
Document "services" field

pchico83:
I was waiting to document `AttachTo` https://github.com/okteto/okteto/issues/2211, but its implementation is being delayed, and people using `services` is asking about why `services` sn not in the current docs

### DIFF
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -563,21 +563,21 @@ selector:
 ### services ([object], optional)
 
 A list of services that you want to put on developer mode along your your development container.
-The services work just like the development container, with one exception: they won't be able to start an interactive session. 
-The containers defined in this section run in detached mode, but they can also mount your local folders in a particular path.
+The services work just like the development container, with one exception: they won't be able to start an interactive session.
 
 For example, imagine that you have a python-based application with an API and a Worker service. If you define the manifest as shown below, running `okteto up` would give you a remote terminal into the web development container, while synchronizing your changes with both web and worker.
 
 ```yaml
-name: web
-command: ["python", "manage.py", "runserver", "0.0.0.0:8080"]
-sync:
-  - .:/app
-services:
-  - name: worker
-    command: ["celery", "worker", "-A", "myproject.celeryconf", "-Q", "default", "-n", "default@%h"]
+dev:
+  web:
+    command: ["python", "manage.py", "runserver", "0.0.0.0:8080"]
     sync:
       - .:/app
+    services:
+      - name: worker
+        command: ["celery", "worker", "-A", "myproject.celeryconf", "-Q", "default", "-n", "default@%h"]
+        sync:
+          - .:/app
 ```
 
 and both deployments, `web` and `worker` will mount your local folder `.` into the path `/app`.
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -560,6 +560,47 @@ selector:
   app.kubernetes.io/name: vote
 ```
 
+### services ([object], optional)
+
+A list of other containers (other than the main development container) that you want to put on developer mode.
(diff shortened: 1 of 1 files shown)


src/content/reference/manifest.mdx
```diff
@@ -560,6 +560,47 @@ selector:
...
 
+### services ([object], optional)
+
+A list of other containers (other than the main development container) that you want to put on developer mode.
+The difference with the main development container and the ones defined in this section is that the main development container gives you an interactive session, where you can, for example, execute a shell.
+The containers defined in this section run in detached mode, but they can also mount your local folders in a particular path.
+
+For example, to put a Django application on developer mode you could use the following manifest:
+
+```yaml
```
rberrelleza, line 571: this looks like the old manifest format. can we use the new format instead?

src/content/reference/manifest.mdx
```diff
@@ -560,6 +560,47 @@ selector:
   app.kubernetes.io/name: vote
 ```
 
+### services ([object], optional)
+
+A list of other containers (other than the main development container) that you want to put on developer mode.
```
rberrelleza, line 565: If the key is called `services`, I think we need to call them `services`in the description no? 

```suggestion
A list of services that you want to put on developer mode along your your development container.
```

src/content/reference/manifest.mdx
```diff
@@ -560,6 +560,47 @@ selector:
...
+The difference with the main development container and the ones defined in this section is that the main development container gives you an interactive session, where you can, for example, execute a shell.
```
rberrelleza, line 566:
```suggestion
The services work just like the development container, with one exception: they won't be able to start an interactive session. 
```

src/content/reference/manifest.mdx
```diff
@@ -560,6 +560,47 @@ selector:
...
+The containers defined in this section run in detached mode, but they can also mount your local folders in a particular path.
```
rberrelleza, line 567: I don't think we should say "detached mode" unless we have a definition for that somewhere. I think it's better to remove this paragraph.

src/content/reference/manifest.mdx
```diff
@@ -560,6 +560,47 @@ selector:
...
+
+For example, to put a Django application on developer mode you could use the following manifest:
```
rberrelleza, line 569: Most people wont'be familiar with the Django application model, I think you need to explain it a bit more. 

Maybe something like this?

```suggestion
For example, imagine that you have a python-based application with an API and a Worker service. If you define the manifest as shown below, running `okteto up` would give you a remote terminal into the web development container, while synchronizing your changes with both web and worker.
```

src/content/reference/manifest.mdx
rberrelleza: I thikn putting the entire manifest (or at least the entire `dev` key0) is more clear.

src/content/reference/manifest.mdx
```diff
@@ -560,6 +560,47 @@ selector:
   app.kubernetes.io/name: vote
 ```
 
+### services ([object], optional)
```
rberrelleza, line 563: In the CLI 2.0, why do we need services, if you can do okteto up to multiple things at the same time? wouldn't it be easier for our customers to support `okteto up web worker`? (I'm mostly curious about the plan and how we differentiate one from the other for our customers)

pchico83: we don't support `okteto up web worker`. We did something similar with the `--detach` flag, but it's only available in the docker-desktop feature branch.
At some point we should have a better experience to activate development on several containers at the same time, but while we get there, `services` is only alternative to cover this scenario today

src/content/reference/manifest.mdx
pchico83: Right, I was copy/pasting too fast :-)

rberrelleza (approved):

PR #160, opened 2022-07-27
Document `dev.okteto.com/skip-snapshot-if-same-namespace: "true"` ann…

pchico83:
…otation

This is useful to avoid cloning data when redeploying the volumes in the namespace where the source snapshot is defined


RinkiyaKeDad (approved):

rberrelleza:
@jmacelroy ie this included in 0.15?

pchico83:
@rberrelleza yes, it is included in 0.15

PR #163, opened 2022-08-02
docs: update supported kubenetes versions

irespaldiza:


RinkiyaKeDad (approved):

src/content/enterprise/aks.mdx
```diff
@@ -28,9 +28,10 @@ This guide will assume that your domain is registered in [AzureDNS](https://docs
 ### Deploy a Kubernetes cluster
 
 We recommend that you follow Azure's [cluster creation guide](https://docs.microsoft.com/en-us/azure/aks/kubernetes-walkthrough).
+Okteto supports Kubernetes versions 1.20 through 1.22. Support for 1.19 is deprecated.
```
rberrelleza, line 31: Remove the deprecated part. I think is clear enough by saying we support X to Y.

rberrelleza (commented):
Can we also add it in the latest release notes that we're dropping support for 1.19?

src/content/enterprise/aks.mdx
irespaldiza: although it is marked for deprecation we support v1.19. From what I understand the change would be:
```
Okteto supports Kubernetes versions 1.19 through 1.22.
```
It is right?

rberrelleza: I thought we no longer support 1.19. @maroshii ?

maroshii: There's confusion about 1.19 support. We are actually taking it into account in the the 0.15.0 testing plan but marked it as deprecated in the release notes.
+1 for "officially" supporting 1.20 1.21 1.22 only (in the docs) and internally supporting 1.19

rberrelleza (approved):

rberrelleza (approved):

src/content/enterprise/eks.mdx
```diff
@@ -28,9 +28,10 @@ This guide will assume that your domain is registered in [Amazon Route53](https:
 ### Deploy a Kubernetes cluster
 
 We recommend that you follow Amazon's [cluster creation guide](https://docs.aws.amazon.com/eks/latest/userguide/create-cluster.html).
+Okteto supports Kubernetes versions 1.20 through 1.22.
```
rberrelleza, line 31: @pchico83 do we want to say 1.21 given the issue with volume snapshots?

pchico83: @rberrelleza we have tested the rest of the functionality and everything was green. Also, I think it's not about the kubernetes version. It's related to the CSI driver version that you install in the EKS cluster.
Therefore, I would keep 1.22 as supported

PR #195, opened 2022-09-14
Update k8s version

irespaldiza:
Do not merge this PR until release the okteto helm chart v1.0


rberrelleza:
Can you add it to the 1.0 branch once it's approved?

src/content/enterprise/aks.mdx
```diff
@@ -28,10 +28,10 @@ This guide will assume that your domain is registered in [AzureDNS](https://docs
 ### Deploy a Kubernetes cluster
 
 We recommend that you follow Azure's [cluster creation guide](https://docs.microsoft.com/en-us/azure/aks/kubernetes-walkthrough).
-Okteto supports Kubernetes versions 1.20 through 1.22.
+Okteto supports Kubernetes versions 1.21 through 1.23.
```
rberrelleza, line 31: Checkr and Dafiti are on < 1.21, what is the plan for them @jmacelroy @pchico83 ? Are they aware of this?

pchico83: I will coordinate with them, but I would say they cannot upgrade to 1.0 if they are not in the right kubernetes version

rberrelleza: Those are two of our most developed customers, we can't tell them "upgrade or else". They are on premium support, so we need to talk to them so they are aware of our plan before we publish this.

pchico83: @rberrelleza I think requiring a kubernetes upgrade is the right choice here and what other players (like nginx controller) are doing. Maintaining anything under 1.21 will make it much harder to support 1.25.
I want to coordinate with every self-hosted customer the upgrade to 1.0, to make it clear that there are breaking changes and that k8s >= 1.21 is required. Sounds reasonable to me given the fact this is a mayor release

pchico83: @rberrelleza yes, they are aware of this, but I will sync with them again things get publcished. This change applies to the installation guides of new clusters

pchico83 (approved):

PR #197, opened 2022-09-16
Document userPodAffinity setting

pchico83:
Document how to enabled/disabled user pod affinities, to prefer pods in the same namespace to be deployed in the same node

### DIFF
--- a/src/content/enterprise/administration/configuration.mdx
+++ b/src/content/enterprise/administration/configuration.mdx
@@ -826,6 +826,15 @@ tolerations:
 For example, if you add the label `okteto-node-pool:build` and the taint `okteto-node-pool=build:NoSchedule` to a node, and you
 are using `buildPool: build`, the buildkit and registry pods are deployed to this node.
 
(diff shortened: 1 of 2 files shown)


src/content/enterprise/administration/configuration.mdx
```diff
@@ -826,6 +826,15 @@ tolerations:
 For example, if you add the label `okteto-node-pool:build` and the taint `okteto-node-pool=build:NoSchedule` to a node, and you
 are using `buildPool: build`, the buildkit and registry pods are deployed to this node.
 
+### userPodAffinity
+
+When enabled, Okteto automatically adds an preferred affinity to every pod to lean towards placing pods in the same namespace in the same node (default: `true`).
```
RinkiyaKeDad, line 831:
```suggestion
When enabled Okteto automatically adds a preferred affinity to every pod, which leans towards placing pods in the same namespace and in the same node (default: `true`).
```

RinkiyaKeDad (approved):

PR #200, opened 2022-09-19
Remove requirements for publicly accessible dns domains

pchico83:
This guide covers on-premise installations where s3-like services are not available and applications must be exposed behind an existing ingress controller, including Okteto

### DIFF
--- a/src/content/self-hosted/aks.mdx
+++ b/src/content/self-hosted/aks.mdx
@@ -20,7 +20,7 @@ In order to fully install Okteto, you'll need the following:
 
 ### Subdomain
 
-You'll need access to an internet accessible subdomain to which you can add a wildcard DNS record, such as dev.example.com.
+You'll need access to a subdomain to which you can add a wildcard DNS record, such as dev.example.com.
 
 This guide will assume that your domain is registered in [AzureDNS](https://docs.microsoft.com/en-us/azure/dns/dns-overview). Other DNS services can be used, but aren't covered here.
 
@@ -81,9 +81,8 @@ helm repo update
 
 ### Getting your Okteto License
 
-Okteto is free for small teams. You get all the features of [Okteto](/free-trial/) for up to 3 users with 3 namespaces each.
-
-Want to use Okteto with a bigger team? [Let's talk](https://okteto.com/#talktous)
+Okteto is free for up to 3 users without a license.
+You can also sign on for the [free trial](/free-trial/) to give access up to 100 users for a month.
 
 ## Installing Okteto
 
--- a/src/content/self-hosted/civo.mdx
+++ b/src/content/self-hosted/civo.mdx
@@ -21,7 +21,7 @@ In order to fully install Okteto, you'll need the following:
 
 ### Subdomain
 
-You'll need to have access to a internet accessible subdomain to which you can add a wildcard DNS record, such as dev.example.com.
+You'll need access to a subdomain to which you can add a wildcard DNS record, such as dev.example.com.
 
 This guide will assume that your domain is registered in [Civo](https://www.civo.com/account/dns). Other DNS services can be used, but aren't covered here.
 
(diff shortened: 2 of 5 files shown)


src/content/self-hosted/on-premise.mdx
```diff
@@ -0,0 +1,195 @@
...
+You'll need to add the Okteto repository in order to be able to install Okteto:
+
+```console
+helm repo add okteto https://charts.okteto.com
+helm repo update
+```
+
+### Getting your Okteto License
+
+Okteto is free for small teams. You get all the features of [Okteto](/free-trial/) for up to 3 users with 3 namespaces each.
```
rberrelleza, line 86: The free trial is for 100 users. We no longer have the per-namespace limit.

rberrelleza: Free trial is not the same than Starter. starter is the one that's free for up to 3 users without a license.

rberrelleza: In the guides we should encourage the free trial

src/content/self-hosted/on-premise.mdx
```diff
@@ -0,0 +1,195 @@
...
+      pictureKey: picture
+      groupsKey: groups
+
+email: admin@example.com
+license: 1234567890ABCD==
+subdomain: dev.example.com
+cluster:
+  endpoint: "https://your-kubernets-cluster-api-endpoint"
+
+ingress-nginx:
```
rberrelleza, line 127: I don't think this part fits the guide. Nginx Ingress can be installed on-premises.

src/content/self-hosted/on-premise.mdx
```diff
@@ -0,0 +1,195 @@
+---
+title:  On-Premise
```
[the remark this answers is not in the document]
rberrelleza: Instead of on-premises I would pitch this as a guide on how to install okteto without using cloud services

src/content/self-hosted/on-premise.mdx
```diff
@@ -0,0 +1,195 @@
...
+      externalIDKey: nickname
+      nameKey: name
+      emailKey: email
+      pictureKey: picture
+      groupsKey: groups
+
+email: admin@example.com
+license: 1234567890ABCD==
+subdomain: dev.example.com
+cluster:
```
rberrelleza, line 124: We should explain where to get this from

src/content/self-hosted/on-premise.mdx
```diff
@@ -0,0 +1,195 @@
...
+ingress-nginx:
+  enabled: false
+
+registry:
+  storage:
+    filesystem:
+      enabled: true
+      persistence:
+        enabled: true
+        size: 400Gi
```
rberrelleza, line 136: What happens when this runs out of space?

src/content/self-hosted/on-premise.mdx
```diff
@@ -0,0 +1,195 @@
...
+    cloud:
+      enabled: false
+```
+
+And configure the field [ingress](/docs/self-hosted/administration/configuration/#ingress) based on the Ingress Controller running in your cluster.
```
rberrelleza, line 141: If you keep this, Can you add a sample?

src/content/self-hosted/on-premise.mdx
```diff
@@ -0,0 +1,195 @@
...
+```
+
+The output will look something like this:
+
+```
+NAME                             TYPE           CLUSTER-IP   EXTERNAL-IP                          PORT(S)                                     AGE
+okteto-ingress-nginx-controller  LoadBalancer   10.0.7.73    a519c8b3b27f94...aks.microsoft.com   80:30795/TCP,443:32481/TCP,1234:30885/TCP   5m
+```
+
+You'll need to take the `EXTERNAL-IP` address and add it to your DNS for the domain you have chosen to use. In AzureDNS, this is done by creating an `A` record with the name  `*`, pointing to the IP of the Load Balancer.
```
rberrelleza, line 189: What about Buildkit?

rberrelleza:
I don't think that on-premises install means non s3 + custom ingress. I find this mix confusing.

I think that these should be two separate advanced installation guides:
1. How to install Okteto on-premises
2. How to use Okteto with an existing ingress controller


If we want this combo guide for a user or customer, I think is better to have it as a how-to in community.okteto.com.

src/content/self-hosted/on-premise.mdx
pchico83: @rberrelleza could you propose an alternative to this paragraph?
```
Okteto is free for small teams. You get all the features of [Okteto](/free-trial/) for up to 3 users with 3 namespaces each.
```

src/content/self-hosted/on-premise.mdx
pchico83: In this guide, it's exposed as an ingress

PR #206, opened 2022-09-22
Document `cluster.endpoint` for GKE

pchico83:
Recently google changed how they store the cluster endpoint in the default namespace to point to the internal cluster IP. This breaks our logic to infer the cluster public endpoint.

We have to document how to configure this value as we do for other providers

### DIFF
--- a/src/content/self-hosted/gke.mdx
+++ b/src/content/self-hosted/gke.mdx
@@ -119,6 +119,7 @@ Download a copy of the [Okteto GKE configuration file](/docs/self-hosted/gke/con
 - Your `license` (optional)
 - A `subdomain`
 - `clientId` and `clientSecret` of the OAuth 2.0 Client you created
+- Your cluster's API server endpoint
 - The name of the `bucket` you created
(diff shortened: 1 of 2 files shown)


src/content/self-hosted/gke.mdx
```diff
@@ -119,6 +119,7 @@ Download a copy of the [Okteto GKE configuration file](/docs/self-hosted/gke/con
 - Your `license` (optional)
 - A `subdomain`
 - `clientId` and `clientSecret` of the OAuth 2.0 Client you created
+- Your cluster's API server endpoint
```
rberrelleza, line 122: Can we link to a google doc on how to get it?

rberrelleza (approved):

src/content/self-hosted/gke.mdx
pchico83: I cannot find anything. I am going to merge because right now the guide is broken. But I want to add the link later, also for the rest of providers

PR #211, opened 2022-09-23
Fix documentation about compose resources

pchico83:


src/content/reference/docker-compose.mdx
```diff
@@ -309,18 +309,15 @@ scale: 2
...
-    reservations:
-      cpu: 300m
-      memory: 500Mi
-    limits:
-      cpu: 500m
-      memory: 800Mi
+resources:
+  requests:
+    cpu: 300m
+    memory: 500Mi
```
rberrelleza, line 315: I don't think is clear where in the compose this goes. Could you add a simple compose instead of the fragment?

PR #215, opened 2022-09-28
Add explanation about using ":" or "@" for componing the OKTETO_BUILD…

pchico83:
…_XXX envvars

Many helm charts have different fields to define the registry, repo and tag of a given deployment. This is way we have the following ennvars:

- `${OKTETO_BUILD_<service>_TAG}`: SHA for the last image build at the registry
- `${OKTETO_BUILD_<service>_REGISTRY}`: the registry URL where the image was pushed
- `${OKTETO_BUILD_<service>_REPOSITORY}`: the name of the image that was pushed (`namespace/_<service>`)

It's common in helm that the image of the deployment is specified like this:

```
image: {{ .Values.registry}}/{{.Values.repository}}:{{.Values.tag}}
```
but this will fail because after `:` you need to specify the tag name, not the tag sha. In order to specify the sha you need to use `@` instead of `:`. In order to fix this, you can set `.Values.tag` to `okteto@${OKTETO_BUILD_<service>_TAG}`.

### DIFF
--- a/src/content/cloud/okteto-cli.mdx
+++ b/src/content/cloud/okteto-cli.mdx
@@ -76,6 +76,11 @@ In order to refer to images in the `build` section of the Okteto Manifest, you c
 - `${OKTETO_BUILD_<service>_REPOSITORY}`: the name of the image that was pushed (`namespace/_<service>`)
 - `${OKTETO_BUILD_<service>_IMAGE}`: the full image reference (using its SHA)
 
+`${OKTETO_BUILD_<service>_IMAGE}` is equivalent to any of the following:
+- `${OKTETO_BUILD_<service>_REGISTRY}`/`${OKTETO_BUILD_<service>_REPOSITORY}`@`${OKTETO_BUILD_<service>_TAG}`
+- `${OKTETO_BUILD_<service>_REGISTRY}`/`${OKTETO_BUILD_<service>_REPOSITORY}`:okteto@`${OKTETO_BUILD_<service>_TAG}`
+
+
 ### Built-in Tools When Deploying To Okteto Cloud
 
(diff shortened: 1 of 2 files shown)


src/content/reference/manifest.mdx
```diff
@@ -63,12 +63,16 @@ Each image supports the following fields:
 
 In order to refer to these images from anywhere in your [Okteto Manifest](/docs/cloud/okteto-cli/#okteto-manifest), you can use the following environment variables:
 
+You can build all these images by running `okteto build`, or `okteto build xxx` to build a single one.
```
RinkiyaKeDad, line 66:
```suggestion
You can build all these images by running `okteto build`, or `okteto build <image-name>` to build a single one.
```

RinkiyaKeDad (approved):

src/content/cloud/okteto-cli.mdx
```diff
@@ -76,6 +76,11 @@ In order to refer to images in the `build` section of the Okteto Manifest, you c
 - `${OKTETO_BUILD_<service>_REPOSITORY}`: the name of the image that was pushed (`namespace/_<service>`)
 - `${OKTETO_BUILD_<service>_IMAGE}`: the full image reference (using its SHA)
 
+`${OKTETO_BUILD_<service>_IMAGE}` is equivalent to any of the following:
```
rberrelleza, line 79: what do we mean by it being equivalent to both values? what determines when I get the first vs getting the one with okteto in the tag?

src/content/reference/manifest.mdx
rberrelleza, line 66:
```suggestion
You can build all the images defined in your Okteto manifest by running `okteto build` or `okteto build <image-name>` to build a single one.
```

pchico83:
Closing this, we will fix it on the CLI so this isn't an issue

PR #218, opened 2022-09-29
Document extraRoleBindings

AgustinRamiroDiaz:
Signed-off-by: Agustin Ramiro Diaz <agustin.ramiro.diaz@gmail.com>

### DIFF
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -517,6 +517,23 @@ The role Okteto assigns to every user. If empty, Okteto will create a default ro
 clusterRole: "role name"
 ```
 
+### extraRoleBindings
+
+Extra `RoleBindings` to be created for every user's service account. This is useful if you want to grant access to other namespace resources in the cluster.  Disabled by default.
+
+The `RoleBindings` will reference `ClusterRoles` and will be scoped to the namespace specified in the configuration. Here's an example:
+
+```yaml
+extraRoleBindings:
+  enabled: true
+  roleBindings:
+    namespace-name1:
+    - cluster-role1
+    - cluster-role2
+    namespace-name2:
+    - cluster-role3
+```
+
 ### convertLoadBalancedServices
 
 Converts services with type LoadBalancer into ClusterIP and automatically creates an ingress. Enabled by default.
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -521,7 +521,7 @@ clusterRole: "role name"
 
 Extra `RoleBindings` to be created for every user's service account. This is useful if you want to grant access to resources in other namespaces of the cluster.  Disabled by default.
 
-The `RoleBindings` will reference `ClusterRoles` and will be scoped to the namespace specified in the configuration. Here's an example:
+The `RoleBindings` will reference `ClusterRoles` and will be scoped to the `namespace` specified in the configuration. Okteto will not create neither the `ClusterRoles` nor the `Namespaces` and expects them to exist in the cluster. Here's an example configuration:
 
 ```yaml
 user: 
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -517,21 +517,22 @@ The role Okteto assigns to every user. If empty, Okteto will create a default ro
 clusterRole: "role name"
 ```
 
-### extraRoleBindings
+### user extraRoleBindings
 
 Extra `RoleBindings` to be created for every user's service account. This is useful if you want to grant access to other namespace resources in the cluster.  Disabled by default.
 
 The `RoleBindings` will reference `ClusterRoles` and will be scoped to the namespace specified in the configuration. Here's an example:
 
 ```yaml
-extraRoleBindings:
-  enabled: true
-  roleBindings:
-    namespace-name1:
-    - cluster-role1
-    - cluster-role2
-    namespace-name2:
-    - cluster-role3
+user: 
+  extraRoleBindings:
+    enabled: true
+    roleBindings:
+      namespace-name1:
+      - cluster-role1
+      - cluster-role2
+      namespace-name2:
+      - cluster-role3
 ```
 
 ### convertLoadBalancedServices
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -519,7 +519,7 @@ clusterRole: "role name"
 
 ### user extraRoleBindings
 
-Extra `RoleBindings` to be created for every user's service account. This is useful if you want to grant access to other namespace resources in the cluster.  Disabled by default.
+Extra `RoleBindings` to be created for every user's service account. This is useful if you want to grant access to resources in other namespaces of the cluster.  Disabled by default.
 
 The `RoleBindings` will reference `ClusterRoles` and will be scoped to the namespace specified in the configuration. Here's an example:
 


src/content/self-hosted/administration/configuration.mdx
```diff
@@ -517,6 +517,23 @@ The role Okteto assigns to every user. If empty, Okteto will create a default ro
...
 ```
 
+### extraRoleBindings
+
+Extra `RoleBindings` to be created for every user's service account. This is useful if you want to grant access to other namespace resources in the cluster.  Disabled by default.
+
+The `RoleBindings` will reference `ClusterRoles` and will be scoped to the namespace specified in the configuration. Here's an example:
+
+```yaml
+extraRoleBindings:
```
pchico83, line 527: @rberrelleza maybe we should move this under `users`? as we did for `serviceAccounts`?

AgustinRamiroDiaz: +1. I was just going to post that. Since we are adding these as new features, it'd be nice to have them organized.

rberrelleza: Yeah, I agree

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -517,6 +517,23 @@ The role Okteto assigns to every user. If empty, Okteto will create a default ro
...
+  enabled: true
+  roleBindings:
```
rberrelleza, line 529: How will this be used? Seems like asking for a existing namespace limits this to very specific scenario. What about giving access to more CRDs, or other types of actions? Is that possible with this?

pchico83: This is to add access to specific namespaces, like `default`. 
Access to CRDs could be cover by:
- Namespaced: https://www.okteto.com/docs/self-hosted/administration/configuration/#clusterrole
- Cluster-wide: https://www.okteto.com/docs/self-hosted/administration/configuration/#globalclusterrole

rberrelleza: > This is to add access to specific namespaces

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -517,6 +517,24 @@ The role Okteto assigns to every user. If empty, Okteto will create a default ro
...
 ```
 
+### user extraRoleBindings
+
+Extra `RoleBindings` to be created for every user's service account. This is useful if you want to grant access to other namespace resources in the cluster.  Disabled by default.
+
+The `RoleBindings` will reference `ClusterRoles` and will be scoped to the namespace specified in the configuration. Here's an example:
+
+```yaml
+user:
```
rberrelleza, line 527: @pchico83 is it `user`or `users`? (just want to make sure it's all the same :P)

src/content/self-hosted/administration/configuration.mdx
rberrelleza: I'm a bit confused with this change, so let's see if I make sense:

1. With this change, you can now bind a series of roleBindings to every account we create. 
2. We already have a setting to assign a single clusterRoleBinding to every account 
3. Why do we call one 'clusterRole' (where we ask the role's name) and call this 'roleBindings'? -> shouldn't we be calling this extraRoles, since we are the ones that create the roleBinding?
4. Why do we allow only one clusterRole but multiple roleBindings? 
5. Could we simplify the model by having a single setting called 'extraRoles', where the user can add a list of cluster and namespace bindings to be applied to the users we create?
6. Are these applied to the SA the installer uses, or only to the one used during 'okteto up'? (or is it the same?)

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -517,6 +517,24 @@ The role Okteto assigns to every user. If empty, Okteto will create a default ro
 clusterRole: "role name"
 ```
 
+### user extraRoleBindings
+
+Extra `RoleBindings` to be created for every user's service account. This is useful if you want to grant access to other namespace resources in the cluster.  Disabled by default.
```
rberrelleza, line 522:
```suggestion
Extra `RoleBindings` to be created for every user's service account. This is useful if you want to grant access to resources in other namespaces of the cluster.  Disabled by default.
```

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -517,6 +517,24 @@ The role Okteto assigns to every user. If empty, Okteto will create a default ro
...
+
+The `RoleBindings` will reference `ClusterRoles` and will be scoped to the namespace specified in the configuration. Here's an example:
+
+```yaml
+user: 
+  extraRoleBindings:
+    enabled: true
+    roleBindings:
+      namespace-name1:
```
rberrelleza, line 531: what happens if the namespace doesn't exist already?

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -517,6 +517,24 @@ The role Okteto assigns to every user. If empty, Okteto will create a default ro
...
+      - cluster-role1
+      - cluster-role2
+      namespace-name2:
+      - cluster-role3
```
rberrelleza, line 535: If this is a cluster-role, why is the setting called `extraRoleBindings` and not `extraClusterRoles`?

rberrelleza:
I'm a bit confused with this change, so let's see if I make sense:

1. We already have a setting to assign a single clusterRoleBinding to every account
2. With this change, you can now bind a series of roleBindings to every account we create.
3. Why do we call one 'clusterRole' (where we ask the role's name) and call this 'roleBindings'? -> shouldn't we be calling this extraRoles, since we are the ones that create the roleBinding? or extraClusterRoles?
4. Why do we allow only one clusterRole but multiple roleBindings?
5. Could we simplify the model by having a single setting called 'extraRoles', where the user can add a list of cluster and namespace bindings to be applied to the users we create?
6. How will this work when we have multiple clusters? 
7. What happens is the role (or namespace) provided in the config doesn't exist? 
8. Are these applied to the SA the installer uses, or only to the one used during 'okteto up'? (or is it the same?)

src/content/self-hosted/administration/configuration.mdx
AgustinRamiroDiaz: As it is now implemented the key is `user`, but I can do the work to replace it with `users`

src/content/self-hosted/administration/configuration.mdx
AgustinRamiroDiaz: This setting is to allow the creation of `RoleBindings` that point to a `ClusterRole`. This is to allow them to get the permissions of the `ClusterRole` but namespaced. 

I get your point about this map not being exactly the rolebindings, since it's a map of `namespace:ClusterRole`. But that wouldn't make it `extraClusterRoles` either. 

This setting is for creating `RoleBindings` so `extraRoleBindings` makes sense, but then the configuration needed is the information we need to create those `RoleBindings`, since we already got the `SA` we only need the `namespace` and `clusterRole`

src/content/self-hosted/administration/configuration.mdx
AgustinRamiroDiaz: As it is now, we will fail with an `internal server error`. We are delegating the `namespaces` and `clusterRoles` management to the admin, since this is a custom feature for them.

We are also handling the `rolebindings` resources at the minimum, always deleting the old ones when the configuration changes.

Another way to handle this could be to create the namespace ourselves or ignore the `roleBinding` if the namespace doesn't exist, but both options that could lead to silent errors

Do you have another behavior in mind?

AgustinRamiroDiaz:
>     1. We already have a setting to assign a single clusterRoleBinding to every account

Yes
 
>     2. With this change, you can now bind a series of roleBindings to every account we create.

Yes. More particularly, roleBindings to clusterRoles in particular namespaces

>     3. Why do we call one 'clusterRole' (where we ask the role's name) and call this 'roleBindings'? -> shouldn't we be calling this extraRoles, since we are the ones that create the roleBinding? or extraClusterRoles?

Makes sense. We can discuss the naming that we find more intuitive for this configurations.
I think that this wouldn't be the crystal clear though, since then the configuration map relates namespace to clusterRoles.
Obviously, when we go for short names we are omitting details, which can lead to confusion and that's why we then have the docs explaining how to use the setting.

>     4. Why do we allow only one clusterRole but multiple roleBindings?

I don't know why is that. Checking out our code it doesn't seem like a tech limitation, so we could have many `clusterRoles`

>     5. Could we simplify the model by having a single setting called 'extraRoles', where the user can add a list of cluster and namespace bindings to be applied to the users we create?

I think we could. I think that if we make it right, it would be a change in existing configuration and would need to be a minor bump.  2 things come to my mind for this setting: common usage and how to distinguish cluster-wide from namespaced

One idea could be to flatten these and assume that if there's no namespace, then it's a clusterRoleBinding
```yaml
extraRoles:
  enabled: true
  roles:
    - namespace: my-ns
      clusterRole: my-cr
    - namespace: my-other-ns
      clusterRole: my-cr
    - clusterRole: this-will-create-clusterRoleBindings
```

>     6. How will this work when we have multiple clusters?

I'm not so sure about this. Maybe @pchico83 or @maroshii have an idea. I know that given how kubernetes works, we won't be able to connect this across clusters with `kubectl`. I'm thinking that we'd need the identity for the user in each cluster so that we are able to asociate this rolebindings in the cluster where the namespace and clusterRole are

>     7. What happens if the role (or namespace) provided in the config doesn't exist?

If the namepsace doesn't exist we fail. I haven't tested the role yet, but I think we don't fail in that case

>     8. Are these applied to the SA the installer uses, or only to the one used during 'okteto up'? (or is it the same?)

Not sure yet. I should research a bit more

pchico83:
@rberrelleza 

```
3. shouldn't we be calling this extraRoles, since we are the ones that create the roleBinding? or extraClusterRoles?
```
I think `extraRoleBindings` is the right term. The user defines new role bindings for each developer. The role binding is created in a given namespace for a given cluster role(s)

```
4. Why do we allow only one clusterRole but multiple roleBindings?
```
We could, but that's a different feature. I wouldn't tight it to this PR

```
5. Could we simplify the model by having a single setting called 'extraRoles', where the user can add a list of cluster and namespace bindings to be applied to the users we create?
```
The behavior and use cases are quite different, having them in a single section will make things harder to explain.

```
8. Are these applied to the SA the installer uses, or only to the one used during 'okteto up'? (or is it the same?)
```
They are the same. Installers run with the user service account

rberrelleza:
> I think extraRoleBindings is the right term. The user defines new role bindings for each developer. The role binding is created in a given namespace for a given cluster role(s)

rberrelleza:
> The behavior and use cases are quite different, having them in a single section will make things harder to explain.

@pchico83 @AgustinRamiroDiaz ok, let's not change the scope of this feature and let's leave it just to specify cluster role/namespace combinations . Can we specify clearly that this requires clusterRole and the namespace to already exist?

rberrelleza:
If this is for cluster-roles, I think the key should be named extraClusterRoles to make it clear that this is to assign extra cluster roles to the users that okteto creates and manages.

pchico83:
> Member

@rberrelleza it's different. `globalClusterRole` creates a `clusterrolebinding` between each user and the `globalClusterRole`. This gives you access to cluster-wide resources.
This feature creates a `rolebinding` between each user, a cluster role, but only in an specific namespace, instead of giving access at the cluster level.
I agree that the three different `roles` could be combined in a single section (or in a better way) but those features are already shipped and used by several customers. I would focus on the feature here, and refactor the old ones as part fo the work we have to improve the helm chart installation and the structure of the different sections

rberrelleza:
> > Member
> 
> @rberrelleza it's different. `globalClusterRole` creates a `clusterrolebinding` between each user and the `globalClusterRole`. This gives you access to cluster-wide resources. This feature creates a `rolebinding` between each user, a cluster role, but only in an specific namespace, instead of giving access at the cluster level. I agree that the three different `roles` could be combined in a single section (or in a better way) but those features are already shipped and used by several customers. I would focus on the feature here, and refactor the old ones as part fo the work we have to improve the helm chart installation and the structure of the different sections

Can you file an issue to fix this in the chart later? right now we only have a handful of users it's easier to deal with this sooner than later. The chart configuration is already super complex, anything we can do to simplify helps :(

AgustinRamiroDiaz:
> > If the namespace doesn't exist we fail. I haven't tested the role yet, but I think we don't fail in that case
> 
> @AgustinRamiroDiaz where does it fail? Can we make sure that we document cleary that we expect the roles to already exist on the cluster?

We access the namespace first and there we create the rolebindings, that's where we fail. 

I'll make sure to document that we expect the namespaces and clusterRoles to exist before hand

src/content/self-hosted/administration/configuration.mdx
rberrelleza: If `user` is the right one, leave it that way 👍

rberrelleza (approved):

PR #225, opened 2022-10-04
Document "forceIngressSubdomain"

pchico83:


src/content/self-hosted/administration/configuration.mdx
```diff
@@ -540,13 +606,23 @@ devStorageClass:
...
+```
+
 ### ingress
 
 Configure default values for the ingress created by Okteto.
 
 - `annotations`: The annotations to apply to all the ingresses created during the Okteto installation.
 - `class`: If set, Okteto will set this as the `ingress.class` of all ingresses managed by Okteto. This is useful if you have more than one ingress controller in your cluster.
 - `forceIngressClass`: If enabled, all ingresses deployed in namespaces managed by Okteto will have the ingress class defined in  `ingress.class` (default: `false`).
+- `forceIngressSubdomain`: If enabled, all ingresses deployed in namespaces managed by Okteto must match the okteto wildcard subdomain (default: `true`).
```
rberrelleza, line 625: Since the section is already called ingress, I think we can call this forceSubdomain to remove the "stuttering"

rberrelleza: I don't think forceSubdomain is a good name for this, since this is affecting the "host" property of the ingress. Maybe we can call it something like "strictHostname" ?

pchico83: I have mixed feelings here. We should do a whole refactor of the `ingress` section, it's mixing several concepts. The mian change is that the configuration for user app ingresses should go under the new `users` section.
But in the meantime, I kind of like it better to follow the same conventions we have for other fields, like `forceIngressClass`. 

In this context, I prefer `forceIngressSubdomain` to `forceIngressHost`, because we don't force the host, we force the subdomain of the host.

rberrelleza: >  follow the same conventions

rberrelleza: > In this context, I prefer forceIngressSubdomain to forceIngressHost, because we don't force the host, we force the subdomain of the host.

AgustinRamiroDiaz: What about leaving the same name for the field but explaining a bit more in the description the behavior around hosts?
```yaml
- `forceIngressSubdomain`: If enabled, the subdomain of the host of all ingresses deployed in namespaces managed by Okteto must match the okteto wildcard subdomain (default: `true`).
```

rberrelleza: @AgustinRamiroDiaz yeah, I think that's a good way of explaining it.

rberrelleza (approved):

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -622,6 +622,7 @@ Configure default values for the ingress created by Okteto.
...
```
pchico83, line 625:
```suggestion
- `forceIngressSubdomain`: If enabled, the subdomain of the host of all ingresses deployed in namespaces managed by Okteto must match the okteto wildcard subdomain (default: `true`).
```

PR #243, opened 2022-11-03
Guide user to use same github app for auth and integration

AgustinRamiroDiaz:
I've also fixed some details on the github.mdx file


src/content/self-hosted/administration/github.mdx
```diff
@@ -21,6 +21,8 @@ There are several reasons for considering this approach including:
 
 ## Configuration Steps 
 
+> If you've already configured GitHub OAuth, you should skip these 2 steps. You'll then use the same Github Application for the Integration as the one from Oauth.
```
rberrelleza, line 24: Isn't it the other way around? The githup app is good for oauth, but not the other way around?

pchico83 (approved):

PR #247, opened 2022-11-10
only have the last 5 versions of our docs on the dropdown

### LINKED ISSUE #248: Only show the last 5 versions of the docs in the dropdown

rberrelleza:
Fixes #248


RinkiyaKeDad (approved):

rberrelleza:
Mmmm this isn't doing what I intended. This is limiting the versions that docusaurus builds. What I want to have all the doc versions available (so we can eventually have an 'archive' page, but only the last 5 on the dropdown selector.

PR #263, opened 2022-12-01
mention the right tiers for custom domains, admin dashboard and others

rberrelleza:


rberrelleza:
@omnomagonz you talked about this at some point. Do you think that https://deploy-preview-263--okteto-docs.netlify.app/docs/cloud/custom-domains/ gives you a good indication that the particular feature is available on:
- Self-Hosted: All tiers
- SaaS:  Scale and Enterprise

omnomagonz:
@rberrelleza I think it's a good start. I'm wondering if maybe something like this could work too:

# Custom Domain Names for your Applications
`[SaaS]` `[Self-Hosted]`
`Scale` `Enterprise`

Where we could use different colors for the product (SaaS, Self-Hosted) and the tiers.

I think we could just ship this change now though and figure out the next iteration if one is warranted 😄

PR #272, opened 2023-01-09
Specify which service accounts are associated to "globalClusterRole"

pchico83:
One of our users was asking about which service account they need on their pods to get access to the "globalClusterRole". This PR addresses that feedback

### DIFF
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -633,7 +633,9 @@ There is only one exception where this storage class is overwritten. In case of 
 
 ### globalClusterRole
 
-Okteto assings this cluster role to every user via a cluster role binding. By default, this behavior is disabled.
+Okteto assings this cluster role to every user via a cluster role binding.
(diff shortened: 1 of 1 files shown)


src/content/self-hosted/administration/configuration.mdx
```diff
@@ -633,7 +633,9 @@ There is only one exception where this storage class is overwritten. In case of
 
 ### globalClusterRole
 
-Okteto assings this cluster role to every user via a cluster role binding. By default, this behavior is disabled.
+Okteto assings this cluster role to every user via a cluster role binding.
+The cluster role binding is created for the user service account in the `okteto` namespace, and the `default` service account on every user namespace.
```
rberrelleza, line 637: I don't understand the part about 'cluster role binding is created'. Do we create it, or do we expect the admin to create it?
```suggestion
The cluster role binding is created for the user service account in the `okteto` namespace, and the `default` service account on every user namespace.
```

pchico83: @rberrelleza the cluster role is created by the okteto admin. the cluster role binding is created by okteto to map the cluster role to each okteto account

PR #282, opened 2023-01-13
Throw build error when Docusaurus finds broken links

huguestennier:
Fix broken links mostly by removing version `> 0.10` from the releases notes

The build will now fail when we have broken links:
https://github.com/okteto/docs/pull/282/files#diff-a038096cbdea434999e1dce5ab497212f1fe18204dde1a027ce3bdd663261a2aR14-R15


rberrelleza:
Yes!

rberrelleza (approved):

PR #284, opened 2023-01-16
added user pod anti-affinity

mnevadom:
related to: https://github.com/okteto/app/issues/5463

### DIFF
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -975,12 +975,12 @@ userDefinedNamespaces: false
 
 When enabled Okteto automatically adds a preferred affinity to every pod, which leans towards placing pods from the same namespace and in the same node (default: `true`).
 
- - `antiaffinity`: When enabled Okteto adds an anti-affinity to the pod to ensure that all pods in the same namespace are placed in the same node. Disabled by default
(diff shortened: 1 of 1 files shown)


src/content/self-hosted/administration/configuration.mdx
```diff
@@ -975,9 +975,13 @@ userDefinedNamespaces: false
 
 When enabled Okteto automatically adds a preferred affinity to every pod, which leans towards placing pods from the same namespace and in the same node (default: `true`).
 
+ - `antiaffinity`: When enabled Okteto adds a pod anti-affinity to ensure all pods in same Namespace are placed together. Disabled by default
```
rberrelleza, line 978:
```suggestion
 - `antiaffinity`: When enabled Okteto adds an anti-affinity to the pod to ensure that all pods in the same namespace are placed in the same node. Disabled by default
```

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -975,9 +975,13 @@ userDefinedNamespaces: false
...
+
 ```yaml
 userPodAffinity:
   enabled: true
+  antiaffinity:
```
rberrelleza, line 983: @pchico83 can we fix the casing to `antiAffinity` so it's consistent with the rest of the properties?

rberrelleza (approved):

Issue #294, opened 2023-01-28
Update all links to use relative paths


rberrelleza:
This is a big one. We need to update ALL the links to use relative rather than absolute paths. Having absolute paths breaks the versioning system, since it always sends you to the root of the docs, rather than to the version.

https://docusaurus.io/docs/markdown-features/links has the guidance we need to follow

PR #295, opened 2023-01-31
Point page links to markdown files

huguestennier:
Fix https://github.com/okteto/docs/issues/294

Since we’re pointing directly to the pages URLs instead of the markdown files, it doesn’t work with versioning as the user will always jump back to the current version

## Test plan

When navigating under a specific version (e.g. `1.1`), you should be able to browser around using links on the page and stay on that version:

https://deploy-preview-295--okteto-docs.netlify.app/docs/1.1/welcome/overview/


src/content/administration/dashboard.mdx
```diff
@@ -61,16 +61,16 @@ You can administer all namespaces in your Okteto instance from the Admin Dashboa
...
 - Get `read-only` access to any namespace managed by Okteto
 
 ## Secrets Section
 
-The Secrets section of the Admin dashboard lets you define cluster-wide secrets. Using the `$SECRET_NAME` syntax, you can refer to these secrets directly in your Okteto manifest. They are also injected into the [environment running](/docs/cloud/okteto-cli/#built-in-tools-when-deploying-to-okteto-cloud) the commands in the `deploy` section of the manifest.
+The Secrets section of the Admin dashboard lets you define cluster-wide secrets. Using the `$SECRET_NAME` syntax, you can refer to these secrets directly in your Okteto manifest. They are also injected into the [environment running](cloud/okteto-cli.mdx#built-in-tools-when-deploying-to-okteto-cloud) the commands in the `deploy` section of the manifest.
 
-> Secrets [defined by the user](/docs/cloud/secrets/) take precedence over the secrets defined here in case of a name collision.
+> Secrets [defined by the user](cloud/secrets.mdx) take precedence over the secrets defined here in case of a name collision.
```
AgustinRamiroDiaz, line 71: Asking to understand for future edits: 
In general we should 
- add the `.mdx` extension
- use path without `/docs/`

src/content/self-hosted/install/releases.mdx
```diff
@@ -63,9 +63,9 @@ November 25, 2022
 
 - Update okteto cli version to [2.9.0](https://github.com/okteto/okteto/releases/tag/2.9.0)
 - Support for self-signed certificates and private CA with Docker
-- Buildkit port is now [configurable](/docs/self-hosted/administration/configuration/#buildkit)
+- Buildkit port is now [configurable](self-hosted/administration/configuration.mdx#buildkit)
```
AgustinRamiroDiaz, line 66: I've been testing these and the links in `releases.mdx` do not change the current version. I think this should be an exception since one could want to check out what was in the 1.2 version and this link should guide them to that version and not the one they are currently on.
Repro steps:
- select 1.4
- go to release notes (this file)
- select a link from the release notes of 1.2 (like this one)
- you'll be redirected to the 1.4 page (what I'm pointing out is that we should redirect to 1.2)

WDYT? cc @rberrelleza @pchico83

src/content/administration/dashboard.mdx
huguestennier: We should point to the mdx file versus the URL. Docusaurus will take care of transforming `cloud/secrets.mdx` to the right URL.

Here’s the official doc: https://docusaurus.io/docs/versioning#link-docs-by-file-paths

rberrelleza: @huguestennier can we point to mdx + an anchor inside the page?

src/content/self-hosted/install/releases.mdx
rberrelleza: Ah, that's an interesting point. At the same time, sending them to another version might be confusing if they are looking at the release notes.  Let me do some research about what other products do.  But let's not block this change due to that, we can iterate on a separate PR

src/content/administration/dashboard.mdx
rberrelleza: Validated that anchors work by clicking on https://github.com/okteto/docs/pull/295/files#diff-cbd2fceab19489448408a5c30e5292108a85b65d5401fe0c5f2213bd8a54a59cR59

rberrelleza (approved):

rberrelleza (approved):

PR #298, opened 2023-02-02
Make it clear how to configure the registry in AWS with IAM Users

pchico83:
The guide to configuring the Registry with IAM Users is confusing. It mentions IAM Roles, but we are not using roles


rberrelleza (approved):
Is this specific to the latest version? If not, can we also do the change in older ones? (I don't think is necessary to do them all. Maybe just 1.0+?)

pchico83:
> Is this specific to the latest version? If not, can we also do the change in older ones? (I don't think is necessary to do them all. Maybe just 1.0+?)

@rberrelleza yes, I was waiting for confirmation about the text and then applying it to all versions. All good with the text?

PR #300, opened 2023-02-02
docs: add icon documentation

jLopezbarb:
Signed-off-by: Javier López Barba <javier@okteto.com>

Add documentation for external resources icons

### DIFF
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -728,11 +728,13 @@ A list of external resources that are part of your development environment.  Use
 external:
   db:
     notes: docs/database.md
+    icon: database
     endpoints:
     - name: db
       url: https://localhost:3306
   functions:
     notes: docs/lambdas.md
+    icon: function
     endpoints:
     - name: data-aggregator
       url: https://fake-id.lambda-url.us-east-1.on.aws
@@ -743,6 +745,18 @@ external:
 #### endpoints ([object], required)
 
 Endpoints contain information on how to access the external resource.
+
+#### icon (string, optional)
+
+ Sets the icon that will be shown in the UI. The supported values for icons are:
+
+ - `container`
+ - `dashboard`
+ - `database`
+ - `default`
+ - `function`
+ - `graph`
+ - `storage`
 
 ##### name (string, required)
 


src/content/reference/manifest.mdx
```diff
@@ -728,16 +738,17 @@ A list of external resources that are part of your development environment.  Use
...
     - name: db
       url: https://localhost:3306
   functions:
     notes: docs/lambdas.md
+    icon: function
     endpoints:
     - name: data-aggregator
       url: https://fake-id.lambda-url.us-east-1.on.aws
     - name: data-processor
-      url: https://fake-id.lambda-url.us-east-2.on.aws
```
rberrelleza, line 740: Do we need to remove this?

rberrelleza (approved):
Which version of the CLI and Okteto requires this?

src/content/reference/manifest.mdx
jLopezbarb: I want to show that url is not required now that we have the dynamic endpoints enabled

jLopezbarb:
> Which version of the CLI and Okteto requires this?

CLI: 2.13 (There is no beta yet)
Okteto chart: 1.5 (There is no beta yet)

src/content/reference/manifest.mdx
```diff
@@ -752,6 +763,20 @@ The name of the endpoint.
 
 The url of the endpoint.
 
+> Dynamic endpoints: The url can be set during deployment stage by sharing `$OKTETO_EXTERNAL_{EXTERNAL_NAME}_ENDPOINTS_{ENDPOINT_NAME}_URL`
+
+#### icon (string, optional)
+
+ Sets the icon that will be shown in the UI. The supported values for icons are:
```
tavferreira, line 770: Although it is self explanatory, should we state that it defaults to default if no icon is set?

src/content/reference/manifest.mdx
AdrianPedriza: I think that removing the url value without specifying why it is not necessary is a bit confusing. Maybe we should add some kind of note, something like: it is not necessary to explicitly indicate the value of a url if it is dynamically generated in the deploy section...?

teresaromero:
this should be pointing to release-1.5 branch?

this would be a link to a new feature under release notes

cc: @jLopezbarb @rberrelleza

viddo:
> > Which version of the CLI and Okteto requires this?
> 
> CLI: 2.13 (There is no beta yet) Okteto chart: 1.5 (There is no beta yet)

How will this behave in v1.5 if the CLI release isn't ready/done yet? 🤔 

@AdrianPedriza do you know? _(pinging your since @jLopezbarb is still on PTO)_

AdrianPedriza:
> > > Which version of the CLI and Okteto requires this?
> > 
> > 
> > CLI: 2.13 (There is no beta yet) Okteto chart: 1.5 (There is no beta yet)
> 
> How will this behave in v1.5 if the CLI release isn't ready/done yet? 🤔
> 
> @AdrianPedriza do you know? _(pinging your since @jLopezbarb is still on PTO)_

simply the validation of the icon value will not be performed at the time of deployment.

viddo:
> simply the validation of the icon value will not be performed at the time of deployment so I guess we should take into account that the user could set a value outside the available ones because the validation will not be found in cli versions <2.13.0.

Aha, thx for prompt reply and clarifying! The frontend app degrades gracefully showing the default icon if given icon doesn't match any of the predefined ones. Given the API returns a `string` type (instead of explicit enums) we're technically adhering to the API contract, so does not sounds like a blocker. 

I am more worried about failing deployments or other errors being caused by not having CLI updated. Sounds like this is fine for now and will work out in next release then soon enough, right?

BTW, @AdrianPedriza can you take care of getting this PR done for v1.5 docs for @teresaromero ? 🙏

AdrianPedriza:
> > simply the validation of the icon value will not be performed at the time of deployment so I guess we should take into account that the user could set a value outside the available ones because the validation will not be found in cli versions <2.13.0.
> 
> Aha, thx for prompt reply and clarifying! The frontend app degrades gracefully showing the default icon if given icon doesn't match any of the predefined ones. Given the API returns a `string` type (instead of explicit enums) we're technically adhering to the API contract, so does not sounds like a blocker.
> 
> I am more worried about failing deployments or other errors being caused by not having CLI updated. Sounds like this is fine for now and will work out in next release then soon enough, right?
> 
> BTW, @AdrianPedriza can you take care of getting this PR done for v1.5 docs for @teresaromero ? 🙏

If I remember correctly the only thing we would not have at this point is a validation of the icon value (it would admit any string), if you cover this case with a default icon... then we are covered. I will create a PR with this changes to `release-1.5`

src/content/reference/manifest.mdx
rberrelleza: I think it's better to add a section on dynamic endpoints, and put that there so it's more obvious.

src/content/reference/manifest.mdx
rberrelleza: agree with @tavferreira

rberrelleza:
Yes @AdrianPedriza @teresaromero this should be merged in the 1.5 branch

src/content/reference/manifest.mdx
```diff
@@ -752,6 +763,20 @@ The name of the endpoint.
 
 The url of the endpoint.
 
+> Dynamic endpoints: The url can be set during deployment stage by sharing `$OKTETO_EXTERNAL_{EXTERNAL_NAME}_ENDPOINTS_{ENDPOINT_NAME}_URL`
```
rberrelleza, line 766: Can we add a sample okteto.yml below this?

```suggestion
##### Dynamic endpoints

The `url` of the endpoint can also be set dynamically during the deployment stage. You can use this for endpoints generated as part of your deployment pipeline (e.g. if you create a Cloud Database). 

After creating the resource, set the URL dynamically by setting an environment variable following the pattern below:

`$OKTETO_EXTERNAL_{EXTERNAL_NAME}_ENDPOINTS_{ENDPOINT_NAME}_URL environment variable`


```# okteto.yaml
deploy:
- name: deploy application
  command: kubectl apply -f  my-app.yml
- name: create database in the cloud
  command: | 
      endpoint_url=$(curl https://internal-service/create-db)
      export OKTETO_EXTERNAL_DB_ENDPOINTS_DASHBOARD_URL=$endpoint_url
external:
  db:
    notes: docs/database.md
    endpoints:
    - name: dashboard```

```

src/content/reference/manifest.mdx
rberrelleza, line 770:
```suggestion
 Sets the icon that will be shown in the UI. The supported values for icons are listed below. If empty, it will default to `default`.
```

src/content/reference/manifest.mdx
```diff
@@ -214,6 +214,16 @@ deploy:
   compose: docker-compose.yml
 ```
 
+You can share environment variables between steps by adding them to `$OKTETO_ENV`:
```
rberrelleza, line 217: Will this work across dependencies? IF so, can we add it to the docs?

AdrianPedriza: it will only works across deploy commands

src/content/reference/manifest.mdx
AdrianPedriza: I added a section about how to use a dynamic endpoint. @rberrelleza can you check it?

Issue #303, opened 2023-02-06
Default the docs to the latest official release of Okteto


rberrelleza:
As we make okteto self-hosted/saas the default experience for the docs (rather than cloud), I would like to change the way we handle versions. 

I want to get our docs to the following state:
1. Navigating to [okteto.com/docs](http://okteto.com/docs) shows you the latest official version of the docs (today, that's 1.4)
2. I want the docs at the root of [github.com/okteto/docs](http://github.com/okteto/docs) to be served on  okteto.com/docs/$NEXT_VERSION (NEXT_VERSION, today, would be 1.15). 
3. I want every released version to have a dedicated URL (like we do today)

This is to enable the team to be able to start doing PRs for the next version of Okteto while it's on development, rather than until it's done. 

Pretty much, it’s the setup that https://docusaurus.io/docs has, but instead of calling the version canary,  I'd like us to call it `1.5` 

I did some experiments with https://github.com/okteto/docs/pull/293 , but couldn't find the magic combo

PR #307, opened 2023-02-10
Default the docs to the latest official release of Okteto

huguestennier:
Fix https://github.com/okteto/docs/issues/303

1. Show `1.4` as the latest version served on `/docs` and show `1.5` as the next version.
2. Removes `welcome/overview` from the homepage URL

### Test plan

These should all work:

https://deploy-preview-307--okteto-docs.netlify.app/docs/
https://deploy-preview-307--okteto-docs.netlify.app/docs/1.5
https://deploy-preview-307--okteto-docs.netlify.app/docs/1.3
https://deploy-preview-307--okteto-docs.netlify.app/docs/1.2

etc


netlify.toml
```diff
@@ -4,23 +4,71 @@
...
-  from = "/docs/:version/*"
-  to = "/docs/:version/welcome/overview/"
+  from = "/docs/0.11/welcome/overview/"
+  to = "/docs/0.11"
   status = 301
 
 [[redirects]]
-  from = "/docs/:version/*"
-  to = "/docs/:version/welcome/overview/"
+  from = "/docs/1.4/*"
```
rberrelleza, line 70: Is this something we'll need to change on every new doc version?

huguestennier: Only this one because we point people at `okteto.com/docs/1.4/XXX` at the moment on the docs site. We’ll be able to remove the redirect when 1.5 points to `okteto.com/docs` and not have to worry about it after

huguestennier: BUT thinking more about it:

Since we use `/docs/1.5/` for the next version, I guess when we point 1.5 to `/docs`, yeah we should update this to redirect `/docs/1.5` to `/docs`

I don’t know how much people bookmarks docs and all, specially in unreleased versions, but it wouldn’t hurt to update it.

huguestennier: BUT thinking more about it:

Since we use `/docs/1.5/` for the next version, I guess when we point 1.5 to `/docs`, yeah we should update this to redirect `/docs/1.5` to `/docs`

I don’t know how much people bookmarks docs and all, specially in unreleased versions, but it wouldn’t hurt to update it.

rberrelleza: would this be part of the "release a new version" task list?

huguestennier: Yep!

rberrelleza (approved):

PR #310, opened 2023-02-12
chore: add build credential helpers

jpf-okteto:
Add example for installer.buildCredentials with IAM Roles for Service Accounts.

### DIFF
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -477,6 +477,18 @@ installer:
   sshSecretName: "okteto-ssh"
 ```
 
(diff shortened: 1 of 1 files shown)


src/content/self-hosted/administration/configuration.mdx
```diff
@@ -477,6 +477,18 @@ installer:
   sshSecretName: "okteto-ssh"
 ```
 
+- `buildCredentialHelpers`: Mapping of registry hostnames as keys to credential helpers as values. Disabled if `privateRegistry` is defined. Registry hostnames must be Fully Qualified Domain Names. Credential helper names must be accessible at installer job pod `$PATH` as `docker-credential-$NAME`.
+
+```yaml
+installer:
+  buildCredentialHelpers:
+    111122223333.dkr.ecr.eu-central-1.amazonaws.com: ecr-login
```
pchico83, line 485: what's `ecr-login`? can it be any string?

PR #318, opened 2023-02-15
Remove reference to dynamic endpoints until 2.13 is released

rberrelleza:


pchico83:
@rberrelleza should we keep it in the "next release"? That's what we do for chart releases, right?

PR #322, opened 2023-02-16
Document the Istio divert driver

pchico83:
We need to create a doc explaining the divert feature, requirements and use cases.
This PR only adds the technical docs to document the fields in the okteto manifest.

### DIFF
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -155,18 +155,9 @@ A list of commands to deploy your development environment.
 
 It's usually a combination of `helm`, `kubectl`, and `okteto` commands.
 
-#### Deploy a Docker Compose file:
+#### Deploy a Helm chart
 
-If you are planning to develop with your `docker-compose` only, you should check out our [Getting Started](using-dev-envs.mdx) guide!
-
-You can also create an `okteto.yml` manifest to get even more out of okteto (e.g. to customize the folders that get synchronized, add extra environment variables, etc...). If you do that, you'll need to update the `deploy` section as follows:
-
-```yaml
-deploy:
-  compose: docker-compose.yml
-```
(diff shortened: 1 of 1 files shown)


versioned_docs/version-1.5/reference/manifest.mdx
```diff
@@ -155,37 +155,43 @@ A list of commands to deploy your development environment.
 
 It's usually a combination of `helm`, `kubectl`, and `okteto` commands.
 
-#### Deploy a Docker Compose file:
+#### Deploy a Helm chart
```
pchico83, line 158: Based on the text above, it reads better if we document helm before docker compose

versioned_docs/version-1.5/reference/manifest.mdx
```diff
@@ -214,16 +223,34 @@ deploy:
...
+    - helm upgrade --install movies chart --set api.image=${OKTETO_BUILD_API_IMAGE} --set frontend.image=${OKTETO_BUILD_FRONTEND_IMAGE}
+  divert:
+    driver: istio
+    namespace: staging
+    service: api
+    virtualServices:
+      - frontend
 ```
 
+When divert is enabled, Okteto will add the header `x-okteto-divert` to every request coming from the developer namespace.
```
rberrelleza, line 244: Can you put the 3 paragraphs of explanation before the sample? I think it reads better that way.

rberrelleza (approved):

rberrelleza:
Does this means that we're ready to fully support divert?

pchico83:
> Does this means that we're ready to fully support divert?

@rberrelleza yes starting from chart 1.6.0. And only for the istio driver. we are still evaluating what to do with the weaver driver

PR #323, opened 2023-02-16
Update mount secrets to use "target" instead of "dst"

pchico83:
Although `dst` keeps working, docker/buildkit updated the docs to use `target`:
https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#run---mounttypesecret


rberrelleza (approved):
do we need to update this on 1.5 as well? (the current version)

pchico83:
> do we need to update this on 1.5 as well? (the current version)

@rberrelleza no need, `dst` also works

PR #333, opened 2023-03-01
Document kubetoken command

AgustinRamiroDiaz:
Documentation for the command added in this pr https://github.com/okteto/okteto/pull/3404

### DIFF
--- a/src/content/reference/cli.mdx
+++ b/src/content/reference/cli.mdx
@@ -357,53 +357,6 @@ $ okteto kubeconfig
  i  Updated kubernetes context 'cloud_okteto_com/cindy' in '/Users/cindy/.kube/config'
 ```
 
-### kubetoken
-
-> This command is only available in okteto clusters
-
-Print Kubernetes cluster credentials in [ExecCredential](https://kubernetes.io/docs/reference/config-api/client-authentication.v1/) format. More information on client side authentication can be found [here](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins).
-
-This command is used internally by okteto to authenticate with the Kubernetes cluster in your kubeconfig, and should only be used when troubleshooting authentication issues.
-
-```console
-$ okteto kubetoken
-```
-
-```json
-{
-  "apiVersion": "client.authentication.k8s.io/v1",
-  "metadata": {
-    "creationTimestamp": null,
-    "managedFields": [
-      {
-        "manager": "backend",
-        "operation": "Update",
-        "apiVersion": "authentication.k8s.io/v1",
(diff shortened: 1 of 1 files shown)


src/content/reference/cli.mdx
```diff
@@ -357,6 +357,53 @@ $ okteto kubeconfig
  i  Updated kubernetes context 'cloud_okteto_com/cindy' in '/Users/cindy/.kube/config'
 ```
 
+### kubetoken
+
+> This command is only available in okteto clusters
+
+Print Kubernetes cluster credentials in [ExecCredential](https://kubernetes.io/docs/reference/config-api/client-authentication.v1/) format. More information on client side authentication can be found [here](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins).
+
+This command is used internally by okteto to authenticate with the Kubernetes cluster in your kubeconfig, and should only be used when troubleshooting authentication issues.
```
rberrelleza, line 366: If it's internal why do we want to expose it to the user? Do we expect them to use this on their own?

If it's just for troubleshooting, could we create a separate section for that, so it's clear that this isn't meant to be use on its own?

rberrelleza: Also, do we have plans to document the troubleshooting aspect of it? (When to use, why, and how?)

src/content/reference/cli.mdx
```diff
@@ -357,6 +357,53 @@ $ okteto kubeconfig
  i  Updated kubernetes context 'cloud_okteto_com/cindy' in '/Users/cindy/.kube/config'
 ```
 
+### kubetoken
+
+> This command is only available in okteto clusters
+
+Print Kubernetes cluster credentials in [ExecCredential](https://kubernetes.io/docs/reference/config-api/client-authentication.v1/) format. More information on client side authentication can be found [here](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins).
```
AdrianPedriza, line 364:
```suggestion
Prints Kubernetes cluster credentials in [ExecCredential](https://kubernetes.io/docs/reference/config-api/client-authentication.v1/) format. More information on client side authentication can be found [here](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins).
```

AgustinRamiroDiaz: I tried following the same tense as the other commands. They are all in infinitive

src/content/reference/cli.mdx
rberrelleza: Yeah. Maybe we can have a section on "troubleshooting" and we can add this as the first one?

rberrelleza (approved):

PR #336, opened 2023-03-06
docs: add release notes for 1.5.1 and 1.4.3

jpf-okteto:
This PR adds release notes for 1.5.1 and 1.4.3, which contains bugfix
for registry authentication when using a custom secret.

### DIFF
--- a/src/content/self-hosted/install/releases.mdx
+++ b/src/content/self-hosted/install/releases.mdx
@@ -5,6 +5,13 @@ sidebar_label: Release Notes
 id: releases
 ---
 
+## 1.5.1
+6 March, 2023
+
+### Bugfixes
+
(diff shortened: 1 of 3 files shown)


ifbyol (approved):
I would also wait until @pchico83 or @rberrelleza reviews it just in case, but it looks good

src/content/self-hosted/install/releases.mdx
```diff
@@ -5,6 +5,13 @@ sidebar_label: Release Notes
 id: releases
 ---
 
+## 1.5.1
+6 March, 2023
+
+### Bugfixes
+
+- Fixed an issue that prevented users from authenticating against the registry when using a custom secret (`.Values.registry.ingress.tlsSecret`).
```
rberrelleza, line 13: Can we call this a "custom TLS secret" or "separate certificate for the registry" so it's more obvious what the use case is?

PR #340, opened 2023-03-09
docs: add publicOverride and privateEndpoints

jpf-okteto:


pchico83 (approved):

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -47,6 +47,26 @@ subdomain: "example.com"
 
 After installation, we recommend that you create a DNS entry for `*.$SUBDOMAIN`, pointing to the public address of your load balancer.
 
+### publicOverride
```
rberrelleza, line 50: @omnomagonz @pchico83 since this is something we've never officially supported before, could we take the opportunity to give it a more standard and self-descriptive name to the property?  (eg publicURL).

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -47,6 +47,26 @@ subdomain: "example.com"
...
+
+Use this property to override where Okteto is available. This option replaces `okteto.$SUBDOMAIN` with you [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of choice.
+
+When set, an additional component is deployed for [Okteto Private Endpoints](cloud/private-endpoints.mdx) feature. You can configure its settings at [`privateEndpoints`] section.
```
rberrelleza, line 54: What settings can be configured? How is this different from when not using publicOverride?

src/content/self-hosted/administration/configuration.mdx
rberrelleza, line 50: Can we have a guide somewhere on how to use this? Changing the publicOverride involves also setting up a new certificate right? Is that only for the main entry point or does it also apply to registry, buildkit, etc?

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -47,6 +47,26 @@ subdomain: "example.com"
...
+
+Authentication callback URL and origin URL will change to the following:
```
rberrelleza, line 56: Will the buildkit and registry URLs change?

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -47,6 +47,26 @@ subdomain: "example.com"
 
 After installation, we recommend that you create a DNS entry for `*.$SUBDOMAIN`, pointing to the public address of your load balancer.
 
+### publicOverride
+
+Use this property to override where Okteto is available. This option replaces `okteto.$SUBDOMAIN` with you [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of choice.
```
rberrelleza, line 52: @jpf-okteto @pchico83 we also need to create a certificate for this FQDN right? can we include instructions on that?

PR #341, opened 2023-03-10
Add a note for git submodules when using GitLab preview environments

pchico83:
Users need to use the git SSH URL of their repos in order to make it work with git submodules.
We should fix this at the product level, but in the meantime, we can document it as a recommendation

### DIFF
--- a/src/content/cloud/preview-environments/gitlab.mdx
+++ b/src/content/cloud/preview-environments/gitlab.mdx
@@ -101,6 +101,7 @@ A few recommendations when creating your preview environment:
 - Create a preview environment per branch/merge request to keep things isolated. We use both `$CI_BUILD_REF_NAME` and `$OKTETO_USERNAME` in the name to ensure the namespace is unique and that we only create one per branch.
 - Pass the URL of your preview environment using the `environment.url` key. That way, the reviewers can directly go to the preview environment from GitLab.
 - Delete both the preview environment, and the namespace when the branch is deleted to cut down on manual deletion.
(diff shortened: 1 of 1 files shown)


src/content/cloud/preview-environments/gitlab.mdx
```diff
@@ -101,6 +101,7 @@ A few recommendations when creating your preview environment:
 - Create a preview environment per branch/merge request to keep things isolated. We use both `$CI_BUILD_REF_NAME` and `$OKTETO_USERNAME` in the name to ensure the namespace is unique and that we only create one per branch.
 - Pass the URL of your preview environment using the `environment.url` key. That way, the reviewers can directly go to the preview environment from GitLab.
 - Delete both the preview environment, and the namespace when the branch is deleted to cut down on manual deletion.
+- Overwrite `CI_REPOSITORY_URL` with the git SSH URL of your repo, and configure [SSH access to your private repos](administration/private-repositories/ssh-key.mdx). This is especially relevant if your repo has private git submodules.
```
rberrelleza, line 104: Is this needed all the time, or just for special cases? If it's for special cases, let's put it in a separate section. 

CI_REPOSITORY_URL is a variable provided by gitlab, how can you override it?

pchico83: @rberrelleza it's always needed. We should fix it at the product level too
Also, as part of your pipeline, you can do `export CI_REPOSITORY_URL=xxx` to overwrite its value

pchico83: If the main repo is private

rberrelleza: Then add that as a subsection. And it would be useful to add a code sample so it's clear what we mean by it.

Or maybe we can use a different env var? @davidpthomas maybe knows about this?

PR #343, opened 2023-03-13
docs: add info about ingress-nginx values/class

jpf-okteto:


src/content/self-hosted/administration/configuration.mdx
```diff
@@ -1266,3 +1266,29 @@ ingress-nginx:
 ```
 
 The full list of values is [available here](https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/values.yaml).
+
+#### Okteto ingress-nginx default values
+
+Okteto Self Hosted applies some defaults to ingress-nginx, which can be checked before overriding them with the following command:
```
rberrelleza, line 1272: can we be specific on the defaults that are applied?  Is this for all ingress-nginx, or just the one installed by us?

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -1266,3 +1266,29 @@ ingress-nginx:
...
+
+```
+helm get values okteto/okteto --jsonpath '{.ingress-nginx}'
+```
+
+#### Use a different ingressClass
+
+Okteto Self Hosted uses the following ingressClass value by default in both, control plane and dev user ingresses: `okteto-nginx`.
```
rberrelleza, line 1280: What do we mean by "control plane" and "dev user" ingresses? Do we use those terms anywhere else in the docs?

```suggestion
Okteto Self Hosted uses the following ingressClass value by default for control plane and dev user ingresses: `okteto-nginx`.
```

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -1266,3 +1266,29 @@ ingress-nginx:
...
+
+To use a different value, use the following snippet:
```
rberrelleza, line 1282: Why would you need to do this? Could we include some info on the use case or pain point this solves?

pchico83: I think using a different ingress class/ingress controller requires its own doc. It also requires configuring other things to make it work. I would remove this section for now

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -1266,3 +1266,29 @@ ingress-nginx:
 ```
 
 The full list of values is [available here](https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/values.yaml).
+
+#### Okteto ingress-nginx default values
+
+Okteto Self Hosted sets specific values on top of ingress-nginx default chart values, to enable some features dependant on the ingress-controller. These values can be checked with the following command:
```
rberrelleza, line 1272:
```suggestion
Okteto sets specific values on the embedded ingress-nginx chart to enable features dependant on the ingress-controller. The values can be checked with the following command:
```

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -1266,3 +1266,29 @@ ingress-nginx:
...
+
+```
+helm get values okteto/okteto --jsonpath '{.ingress-nginx}'
+```
+
+#### Use a different ingressClass
+
+Okteto Self Hosted uses the following ingressClass value by default for all ingresses managed by Okteto: `okteto-nginx`.
```
rberrelleza, line 1280:
```suggestion
Okteto uses the following ingressClass value by default for all ingresses managed by Okteto: `okteto-nginx`.
```

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -1266,3 +1266,29 @@ ingress-nginx:
...
+
+#### Okteto ingress-nginx default values
+
+Okteto Self Hosted sets specific values on top of ingress-nginx default chart values, to enable some features dependant on the ingress-controller. These values can be checked with the following command:
+
+```
+helm get values okteto/okteto --jsonpath '{.ingress-nginx}'
+```
+
+#### Use a different ingressClass
```
rberrelleza, line 1278:
```suggestion
#### Customize the `ingressClass` used by Okteto
```

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -1266,3 +1266,29 @@ ingress-nginx:
...
+
+Okteto Self Hosted uses the following ingressClass value by default for all ingresses managed by Okteto: `okteto-nginx`.
+
+To use a different value, use the following snippet:
```
rberrelleza, line 1282:
```suggestion
In order to have Okteto set a different ingressClass to its ingress, use the following configuration snippet in your Helm values files.
```

PR #349, opened 2023-03-15
Add root path redirect

huguestennier:
Redirects `/` to `/docs` so if you try to access the deploy preview URL, you won’t see a 404

### Test plan

https://deploy-preview-349--okteto-docs.netlify.app/ should redirect to https://deploy-preview-349--okteto-docs.netlify.app/docs

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -68,6 +68,11 @@
   status = 301
 
 [[redirects]]
+  from = "/"
+  to = "/docs"
+  status = 301
+
+[[redirects]]
   from = "/*"
   to = "/docs/404.html"
   status = 404


rberrelleza (approved):

rberrelleza:
Is there a way to also add this so that, when doing okteto up, `/`redirects to `/docs`?

huguestennier:
> Is there a way to also add this so that, when doing okteto up, /redirects to /docs?

Hmm, is there a way to set a base directory with Okteto?

Also, shouldn’t it follow the `npm start` command?

Screenshot 2023-03-15 at 7 30 34 PM

PR #339, opened 2023-03-09
Release notes for release 1.6

ifbyol:
Release notes for `1.6`

There is a couple of things left:
* Documentation for 2 features: The default ingress class we have changed in `1.6` and the public override public availability and the changes related to the private endpoints handler. @jpf-okteto is working on them. The idea would be to include a link to those documents in the corresponding feature point
* CVE fixes included in the release. There were several but I'm not sure which ones. @jpf-okteto once you finish the other documentation tasks, can you include those points in the release notes?

PR #350, opened 2023-03-16
Proposal for removing kubernetes supported versions for public guides

ifbyol:
This PR come from a comment in the PR for the release notes for `1.6.0` version: https://github.com/okteto/docs/pull/339#discussion_r1134394072

More context: In one nexus we commented about changing the supported Kubernetes version in our docs and we realized we have that information "duplicated" in several places. We commented about leaving that information only in the preparation page and link it in the public guides. Also, the same for the technical specs. In fact, we already have some discrepancy there because in the preparation guide we mention `100GB` for the storage but `250GB` in the providers public guides. So, with that purpose, we included those changes.

Finally, those changes were reverted from the release notes PRs and I opened this PR to have that conversation here. 

In this case, I just removed from the public guides the Kubernetes supported versions but I left the technical specs and @rberrelleza  had a point on mentioning that tech spec could mean different things in each Cloud Provider.

Anyways, I opened the PR to follow up the PR comment but I know there will be some changes in the public guides and we might close this one and include it in those improvements

### DIFF
--- a/src/content/self-hosted/digitalocean.mdx
+++ b/src/content/self-hosted/digitalocean.mdx
@@ -29,8 +29,7 @@ In the rest of this guide, we will refer to this subdomain as `DOMAIN`.
 
 ### Deploy a Kubernetes cluster
 
-If you are not familiar with this step, we recommend that you follow DigitalOcean's [cluster creation guide](https://www.digitalocean.com/docs/kubernetes/how-to/create-clusters/).
-Okteto supports Kubernetes versions 1.22 through 1.24.
+If you are not familiar with this step, we recommend that you follow DigitalOcean's [cluster creation guide](https://www.digitalocean.com/docs/kubernetes/how-to/create-clusters/) and check our [preparation guide](self-hosted/install/preparation.mdx#deploy-a-kubernetes-cluster).
 
 To get started with Okteto, follow these specs:
 
--- a/src/content/self-hosted/install/preparation.mdx
+++ b/src/content/self-hosted/install/preparation.mdx
@@ -36,8 +36,8 @@ Okteto supports Kubernetes versions 1.22 through 1.24.
 
 We recommend the following specs:
 - v1.23
-- A pool with at least 3 nodes with a minimum of 4CPUs and 16 GB of Memory
-- 100 GB per disk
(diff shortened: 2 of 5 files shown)


src/content/self-hosted/digitalocean.mdx
```diff
@@ -29,8 +29,7 @@ In the rest of this guide, we will refer to this subdomain as `DOMAIN`.
 
 ### Deploy a Kubernetes cluster
 
-If you are not familiar with this step, we recommend that you follow DigitalOcean's [cluster creation guide](https://www.digitalocean.com/docs/kubernetes/how-to/create-clusters/).
-Okteto supports Kubernetes versions 1.22 through 1.24.
```
rberrelleza, line 33: I don't think we should remove this. We want this guide to be a single doc the installer can follow, not to have to jump back and forth.

src/content/self-hosted/install/preparation.mdx
```diff
@@ -36,8 +36,8 @@ Okteto supports Kubernetes versions 1.22 through 1.24.
 
 We recommend the following specs:
 - v1.23
-- A pool with at least 3 nodes with a minimum of 4CPUs and 16 GB of Memory
-- 100 GB per disk
+- A pool with at least 3 nodes with a minimum of 4CPUs and 32 GB of Memory
```
rberrelleza, line 39: Per node?

src/content/self-hosted/install/preparation.mdx
```diff
@@ -36,8 +36,8 @@ Okteto supports Kubernetes versions 1.22 through 1.24.
...
+- 250 GB per disk
```
rberrelleza, line 40: Is this on every node? What kind if storage?

src/content/self-hosted/digitalocean.mdx
rberrelleza: I think we should also add a recommendation on which type of instance we recommend for every cloud provider, using the specific cloud nomenclature.

ifbyol:
I will close this PR as this was just came up during a daily and it seems it's not what we want. We can always reopen it if needed

PR #353, opened 2023-03-17
Describe aws/gcp actions to configure IAM roles/Workload ID for the O…

pchico83:
…kteto Registry.

We can assume the user will know how to attach a policy, to a Kubernetes service account, but it's good to mention in case the user forgets to do it. I have already troubleshoot two cases where the user was missing that part.

### DIFF
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -64,7 +64,11 @@ You can overwrite Okteto's public URL using the [publicOverride](self-hosted/adm
 
 ### publicOverride
 
-Use this property to override where Okteto is available. This option replaces `okteto.$SUBDOMAIN` with you [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of choice.
+Use this property to override the Public URL where Okteto is available. This option replaces `okteto.$SUBDOMAIN` with your [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of choice.
+
+```yaml
+publicOverride: "example.com"
+```
 
 When set, an additional component is deployed for [Okteto Private Endpoints](cloud/private-endpoints.mdx) feature.
 You can configure its settings using the [privateEndpoints](self-hosted/administration/configuration.mdx#privateEndpoints) section.
@@ -452,12 +456,21 @@ installer:
       memory: 50Mi
 ```
 
-- `buildCredentialHelpers`: Mapping of registry hostnames as keys to credential helpers as values. Disabled if `privateRegistry` is defined. Registry hostnames must be Fully Qualified Domain Names. Credential helper names must be accessible at installer job pod `$PATH` as `docker-credential-$NAME`.
+#### buildCredentialHelpers
+
+Configure IAM roles to pull private images from your Dockerfiles. Disabled if [privateRegistry](self-hosted/administration/configuration.mdx#privateregistry) is defined.
+
+For private ECRs, add a mapping of registry hostnames as keys to credential helpers as values. Registry hostnames must be Fully Qualified Domain Names.
 
 ```yaml
 installer:
   buildCredentialHelpers:
     111122223333.dkr.ecr.eu-central-1.amazonaws.com: ecr-login
+```
+
+Create an IAM role with access to your private ECR, assume this role by the audience of Okteto Developer Service Accounts (service accounts in the `okteto` namespace) and apply these changes to the Okteto configuration:
+
+```yaml
 user:
   serviceAccount:
     annotations:
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -456,27 +456,6 @@ installer:
       memory: 50Mi
 ```
 
-#### buildCredentialHelpers
-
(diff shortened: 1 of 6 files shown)


src/content/self-hosted/administration/configuration.mdx
```diff
@@ -49,7 +49,11 @@ After installation, we recommend that you create a DNS entry for `*.$SUBDOMAIN`,
 
 ### publicOverride
 
-Use this property to override where Okteto is available. This option replaces `okteto.$SUBDOMAIN` with you [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of choice.
+Use this property to override the Public URL where Okteto is available. This option replaces `okteto.$SUBDOMAIN` with your [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of choice.
```
rberrelleza, line 52:
```suggestion
Use this property to override the Public URL where Okteto is available. This option replaces `okteto.$SUBDOMAIN` with the [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) you provide.
```

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -261,6 +265,8 @@ cloud:
 
 ##### IAM Role for Service Account (IRSA)
 
+Create a IAM role with access to the Okteto Registry bucket, bind the IAM role to the Okteto Registry Kubernetes service account and apply these changes to the Okteto configuration:
```
rberrelleza, line 268: Can we link to official docs on this?

rberrelleza: Also, can we specify the exact permissions that we require for the IAM?

rberrelleza, line 268: Can we also tell the reader the exact name of the service account? Or show them how to get it?

rberrelleza, line 268:
```suggestion
Create a IAM role with access to the Okteto Registry bucket, bind the IAM role to the Okteto Registry Kubernetes service account and apply the changes below to your Helm configuration file:
```

rberrelleza: We also need to tell them to upgrade no?

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -568,12 +577,21 @@ installer:
       memory: 50Mi
 ```
 
-- `buildCredentialHelpers`: Mapping of registry hostnames as keys to credential helpers as values. Disabled if `privateRegistry` is defined. Registry hostnames must be Fully Qualified Domain Names. Credential helper names must be accessible at installer job pod `$PATH` as `docker-credential-$NAME`.
+#### buildCredentialHelpers
+
+Configure IAM roles to pull private images from your Dockerfiles. Disabled if [privateRegistry](self-hosted/administration/configuration.mdx#privateregistry) is defined.
+
+For private ECRs, add a mapping of registry hostnames as keys to credential helpers as values. Registry hostnames must be Fully Qualified Domain Names.
```
rberrelleza, line 584: What's an ECR?

rberrelleza: What are the accepted values for the credential helpers?

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -568,12 +577,21 @@ installer:
...
 
 ```yaml
 installer:
   buildCredentialHelpers:
     111122223333.dkr.ecr.eu-central-1.amazonaws.com: ecr-login
+```
+
+Create an IAM role with access to your private ECR, assume this role by the audience of Okteto Developer Service Accounts (service accounts in the `okteto` namespace) and apply these changes to the Okteto configuration:
```
rberrelleza, line 592: We need to explain how to do this. Is not trivial and it's not that standard yet.

versioned_docs/version-1.6/self-hosted/administration/configuration.mdx
```diff
@@ -49,7 +49,11 @@ After installation, we recommend that you create a DNS entry for `*.$SUBDOMAIN`,
 
 ### publicOverride
 
-Use this property to override where Okteto is available. This option replaces `okteto.$SUBDOMAIN` with you [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of choice.
+Use this property to override the Public URL where Okteto is available. This option replaces `okteto.$SUBDOMAIN` with your [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) of choice.
```
rberrelleza, line 52:
```suggestion
Use this property to override the Public URL where Okteto is available. This option replaces `okteto.$SUBDOMAIN` with the [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) you provided.
```

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -261,6 +265,8 @@ cloud:
 
 ##### IAM Role for Service Account (IRSA)
 
+Create a IAM role with access to the Okteto Registry bucket, bind the IAM role to the Okteto Registry Kubernetes service account and apply the changes below to your Helm configuration file:
```
rberrelleza, line 268: This reads like we are in a rush :P. We need to take our time and explain every step.  We don't have to write everything ourselves, we can send them to the corresponding docs on the cloud provider's doc. But we need to fully document the exact permissions needed for the role, people always get this wrong. 

The same goes for the GCP section.

PR #360, opened 2023-03-27
Update documentation for buildkit

pchico83:
The default installation for buildkit won't use a PVC anymore. The PVC must be enabled explicitly.

This PR covers the following changes related to buildkit:

- Document the new buildkit fields
- Update our cloud provider guides to enable buildkit persistance
- Update the release notes

**Note**: this isn't a breaking change. Existing installation will keep using the existing PVC

@maroshii @teresaromero mention you to approve the tech docs since you have worked on this

### DIFF
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -445,9 +445,14 @@ The build service. It's used in combination with `okteto build` to build contain
 - `podManagementPolicy`: The [podManagementPolicy](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-management-policies) of the buildkit pods. Defaults to `Parallel`.
 - `replicaCount`: The number of buildkit pods. It defaults to 1.
 - `resources`: The resources for the buildkit pods.
-- `storage.class`: The storage class of the volume attached to every buildkit pod to persist the buildkit cache.
-- `storage.size`: The size of the volume attached to every buildkit pod to persist the buildkit cache.
-- `storage.cache`: The size of the buildkit cache to store image caches. It should be 30Gi smaller than `storage.size`.
+- `serviceAccount.create`: create a service account for buildkit. True by default.
+- `serviceAccount.name`: Buildkit service account name. Defaults to `okteto-buildkit`.
+- `serviceAccount.annotations`: Annotations for the buildkit service account.
+- `serviceAccount.labels`: Labels for the buildkit service account.
+- `persistency.enabled`: Configures a persistence volume for buildkit. False by default.
+- `persistency.storageClass`: The storage class of the persistence volume attached to every buildkit pod.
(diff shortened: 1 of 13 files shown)


maroshii (approved):

omnomagonz (approved):

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -445,9 +445,14 @@ The build service. It's used in combination with `okteto build` to build contain
...
-- `storage.size`: The size of the volume attached to every buildkit pod to persist the buildkit cache.
-- `storage.cache`: The size of the buildkit cache to store image caches. It should be 30Gi smaller than `storage.size`.
+- `serviceAccount.create`: create a service account for buildkit. True by default.
+- `serviceAccount.name`: Buildkit service account name. Defaults to `okteto-buildkit`.
+- `serviceAccount.annotations`: Annotations for the buildkit service account.
+- `serviceAccount.labels`: Labels for the buildkit service account.
+- `persistence.enabled`: Configures a persistence volume for buildkit. False by default.
+- `persistence.storageClass`: The storage class of the persistence volume attached to every buildkit pod.
+- `persistence.size`: The size of the persistence volume attached to every buildkit pod.
+- `persistence.cache`: The size of the buildkit cache to store image caches. It should be around 30Gi smaller than `storage.size`.
```
rberrelleza, line 455: can you add the default values?

rberrelleza (approved):
we need to add a 'breaking change' notice in the release notes and in the upgrade guide.

src/content/self-hosted/install/releases.mdx
```diff
@@ -11,6 +11,7 @@ TBD
 ### Features
 
 - Support for the annotation [dev.okteto.com/endpoints](cloud/ssl.mdx#customizing-the-endpoints-shown-in-the-okteto-ui) to customize the endpoints shown in the Okteto UI.
+- Support for configuring the [service account](self-hosted/administration/configuration.mdx#buildkit) of buildkit.
```
omnomagonz, line 14: Do we further define and discuss what a `service account` is and how it fits into the Okteto <> buildkit relationship?

With documentation I tend to lean towards over-communicating to make sure we don't make too many assumptions. But this could also be something our audience (devs) already understand really well.

teresaromero (approved):

src/content/self-hosted/install/releases.mdx
omnomagonz: Yea, that makes sense to me/us 😄 

I'd like for us to take an approach of targeting junior developers in terms of how we talk about things and/or explicitly educate on some things. This _could_ be a way to speak technically enough for our audience but also ensure we're focusing on education and an easy experience in getting Okteto up and running so the first impression is solid.

This can be a non-blocking change now and in the future; just worth considering imo

PR #362, opened 2023-03-27
Document token-based authentication method

pchico83:
Document the token-based authentication and add it to the release notes

### DIFF
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -250,9 +250,7 @@ auth:
     enabled: true
 ```
 
-#### Advanced Cloud Scenarios
-
-It is possible to use separate cloud providers for DNS than for storage if needed. [Reach out to support](https://okteto.com/support), we're always happy to help!
+Follow our post-installation notes to learn how to login to the Okteto UI and configure your Okteto CLI with a randomly generated token.
 
 ## Okteto Components
 
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -235,9 +235,9 @@ The `mapping` fields are optional. Use them to configure the mapping between Okt
 
 #### Token
 
-The token-based authentication method is the default authentication method.
-It allows users to authenticate themselves using a random token of 8 characters.
-This method is suitable for testing scenarios where only one user needs to access Okteto.
+Token-based is the default authentication method.
+This method permits users to validate their identity by employing a randomly generated token comprising eight characters.
+The token-based authentication approach is an optimal choice for testing/evaluation scenarios, where access to Okteto is restricted to a single user.
 
 ```yaml
 auth:
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -235,16 +235,20 @@ The `mapping` fields are optional. Use them to configure the mapping between Okt
 
 #### Token
 
-Token-based is the default authentication method.
-This method permits users to validate their identity by employing a randomly generated token comprising eight characters.
-The token-based authentication approach is an optimal choice for testing/evaluation scenarios, where access to Okteto is restricted to a single user. This is not recommended for multiple users or production setups.
+Okteto defaults to token-based authentication to simplify the installation and configuration of Okteto self-hosted.
+
+This enables you to use a token for a single user, login to the Okteto UI, and start exploring Okteto's features and functionality much faster than configuring other authentication methods.
+Other authentication methods are more time-consuming and cumbersome to configure.
+
+Once you've completed the initial configuration of Okteto using token-based authentication and explored the product, you can then switch to another authentication method.
+Token-based authentication would no longer be ideal as you expand to multiple users and incorporating production environments into Okteto.
+
 
 ```yaml
 auth:
   token:
     enabled: true
 ```
-
 
 #### Advanced Cloud Scenarios
 
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -85,7 +85,7 @@ https://private-endpoints.${subdomain}
 
 ### auth
 
-Okteto supports using Bitbucket, GitHub, Google, or OpenID Connect as authentication providers.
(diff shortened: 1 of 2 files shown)


src/content/self-hosted/administration/configuration.mdx
```diff
@@ -217,6 +217,18 @@ The `mapping` fields are optional. Use them to configure the mapping between Okt
 
 > Your provider needs to support the [UserInfo endpoint](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) in order to be used with Okteto. This authentication option follows the OpenID standard, and it has been validated with Okta, PingIdentity, and GitLab.
 
+#### Token
+
+Token-based is the default authentication method.
+This method permits users to validate their identity by employing a randomly generated token comprising eight characters.
+The token-based authentication approach is an optimal choice for testing/evaluation scenarios, where access to Okteto is restricted to a single user.
```
rberrelleza, line 224:
```suggestion
The token-based authentication approach is an optimal choice for testing/evaluation scenarios, where access to Okteto is restricted to a single user. This is not recommended for multiple users or production setups.
```

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -217,6 +217,18 @@ The `mapping` fields are optional. Use them to configure the mapping between Okt
...
+
+```yaml
+auth:
+  token:
+    enabled: true
```
rberrelleza, line 229: we need to add instructions on how to retrieve the token (or pass a pre-existing value, if we support that)

src/content/self-hosted/administration/configuration.mdx
```diff
@@ -69,7 +69,7 @@ https://private-endpoints.${subdomain}
 
 ### auth
 
-Okteto supports using Bitbucket, GitHub, Google, or OpenID Connect as authentication providers.
+Okteto supports using Bitbucket, GitHub, Google, OpenID Connect, or Token-based as authentication providers.
```
omnomagonz, line 72: Small suggested change:

> or token-based methods as authentication

Something too add a bit of clarity here since grammatically this seems a bit odd to me 😄 "Methods" can be a different description if we like, but I'm assuming that's the intent here.

src/content/self-hosted/install/releases.mdx
```diff
@@ -11,6 +11,7 @@ TBD
 ### Features
 
 - Support for the annotation [dev.okteto.com/endpoints](cloud/ssl.mdx#customizing-the-endpoints-shown-in-the-okteto-ui) to customize the endpoints shown in the Okteto UI.
+- Support for [token-based authentication](self-hosted/administration/configuration.mdx#token) to facilitate testing scenarios.
```
omnomagonz, line 14: Maybe an obvious question: why is token-based auth specifically good for facilitating test scenarios? My assumption is that auth providers don't really have a good experience for test scenarios so token-based auth allows devs to emulate that process. Is it worth articulating this (or the correct rationale) in docs for better clarity?

I'm thinking about people not familiar with Okteto already, devs who maybe aren't familiar with this nuance, and/or maybe more junior devs who just don't understand the dynamics yet.

src/content/self-hosted/administration/configuration.mdx
pchico83: This is what we show after `helm install`:
```
1. Create the following DNS record, pointing it to the NGINX controller service External-IP:
    "*.pablo.dev.okteto.net"

You can retrieve the External IP by running:
    kubectl get service -l=app.kubernetes.io/name=ingress-nginx,app.kubernetes.io/component=controller --namespace=okteto

2. Create the following DNS record, pointing it to the Buildkit service External-IP:
    "buildkit.pablo.dev.okteto.net"

You can retrieve the External IP by running:
    kubectl get service -l=app.kubernetes.io/instance==pablo,app.kubernetes.io/component=buildkit --namespace=okteto

3. Once you create both DNS entries you can access your Okteto instance at this URL:
    https://okteto.pablo.dev.okteto.net/#/login?token=d9f2fd6c

In order to use the Okteto CLI with the instance, you need to execute the following commands:
- Install the Okteto CLI:
    curl https://get.okteto.com -sSfL | sh

- Configure the context of your Okteto CLI
    okteto context use https://okteto.pablo.dev.okteto.net --token d9f2fd6c
``` 

Would you show that output? or would you put it here with different wording?

src/content/self-hosted/install/releases.mdx
pchico83: Configuring other auth methods is more time-consuming and it usually requires approvals from other teams. What I mean here is, that if you want to give it a try to okteto, it's way easier to configure a token for a single user, login to the Okteto UI, and get a sense of our platform. Then, if you like it, it's recommended to use another auth method, which are more secure and they also support multiuser. Token-based is to make it easier to experiment with the Okteto Platform.
How would you phrase that?

omnomagonz: @pchico83 I think that's a perfect explanation of why we're adding this 👍 Maybe we can add something like that to your proposed `Token` section in `/self-hosted/administration/configuration.mdx#token`. My proposal:

Current language you have in this commit:

> Token-based is the default authentication method.
> This method permits users to validate their identity by employing a randomly generated token comprising eight characters.
> The token-based authentication approach is an optimal choice for testing/evaluation scenarios, where access to Okteto is restricted to a single user. This is not recommended for multiple users or production setups.

My suggestion:

> Okteto defaults to token-based authentication to simplify the installation and configuration of Okteto self-hosted.
> 
> This enables you to configure a token for a single user, log in to the Okteto UI, and start exploring Okteto's features and functionality much faster than configuring IAM tools. Configuring other authentication tools typically requires approvals from other teams and are more time-consuming and cumbersome to configure.
> 
> Once you've completed the initial configuration of Okteto using token-based authentication and explored the product, you can then switch to a more optimal IAM solution. Token-based authentication would no longer be ideal as you expand to multiple users and incorporating production environments into Okteto.

WDYT? This way the documentation can be a bit more direct, educational, and purposeful in explaining why we've introduced a feature 😄

pchico83: That sounds amazing, thanks @omnomagonz !

src/content/self-hosted/administration/configuration.mdx
pchico83: I added a sentence on this, let me know what think

PR #378, opened 2023-04-27
Document custom labels and annotations in ingress managed by Okteto

pchico83:
I am also removing the `tlsSecret` options because they are broken at the moment and they aren0t used by any user today


src/content/self-hosted/install/releases.mdx
```diff
@@ -8,6 +8,9 @@ id: releases
 ## 1.8.0
 TBD
 
+### Features
+- Support for custom [labels and annotations](self-hosted/administration/configuration.mdx#namespace) in ingresses managed by Okteto.
+
 ### Breaking changes
 - The default value for the buildkit service session affinity is now `None`. To keep using `ClientIP` as `sessionAffinity` you need to set this value in your [buildkit configuration](self-hosted/administration/configuration.mdx#buildkit).
```
rberrelleza, line 15: What is the purpose of this change?

pchico83: `sessionAffinity` is not supported for every cloud provider. In particular, it's not supported by AWS.
We have a `cloud` section before where the user specified his cloud provider. That doesn't exist anymore.

PR #382, opened 2023-04-29
Move self-hosted configuration to the first level

pchico83:
Self-hosted configuration is one of the most visited paged by our users, but it takes three clicks on the left menu to get there. This PR puts the self-hosted configuration doc one level up

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -111,13 +111,13 @@ module.exports = {
           "self-hosted/gke"
         ]
       },
+      "self-hosted/administration/configuration",
       {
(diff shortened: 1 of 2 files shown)


rberrelleza:
I don't fully get the purpose of this change. If the section is one of the most visited already, why do we need to move it? What is the intended goal of the change since users are not having issues finding it?

pchico83:
@rberrelleza I am not sure of the change either, I wanted to experiment with it in the preview environment.
I am going to close it for now

PR #385, opened 2023-05-03
docs: fix windows download link

jpf-okteto:
Versioned docs at 1.6 and 1.7 contain a wrong Windows Okteto CLI
download link.

Related to:
- https://github.com/okteto/docs/pull/371

### DIFF
--- a/versioned_docs/version-1.6/getting-started.mdx
+++ b/versioned_docs/version-1.6/getting-started.mdx
@@ -37,7 +37,7 @@ $ brew install okteto
 
 ### Windows
 
-Download https://downloads.okteto.com/cli/okteto.exe and add it to your `$PATH`.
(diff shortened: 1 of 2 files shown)


rberrelleza (commented):
nit: Most users don't care about stable/other version, they want the last version that works. Do we need to expose the release type in the URL? why can't we just do downloads.okteto.com/cli/okteto.exe?  cc @omnomagonz

rberrelleza (approved):

omnomagonz:
I'm supportive of your suggestion @rberrelleza.

I've seen the release in the filename, e.g. `okteto-stable.exe` or something (usually for *nix distributions), but I agree having the release be part of the URL seems unnecessary.

The exception I could see is if in some scenarios users wanted to get an older/specific version and placing that in `.com/cli/v2.10/okteto.exe` but even then it seems cleaner to name the file itself (`okteto-cli-2-10.exe`) potentially.

PR #389, opened 2023-05-05
Document upgrade to Kubernetes 1.25 process

pchico83:
Document clean upgrade to Kubernetes 1.25 by disabling pod security policies first


src/content/self-hosted/install/upgrade.mdx
```diff
@@ -154,3 +154,23 @@ helm install cert-manager jetstack/cert-manager --namespace=okteto --version=v1.
 Once installed, cert-manager will continue to renew your existing certificates.
 Please refer to [their docs](https://cert-manager.io/docs/) if you run into any installation or certificate renewal issues.
 
+## Upgrade to Kubernetes 1.25
```
pchico83, line 157: I think this would be better in a dedicated doc that talks about our Kubernetes versions support, but I think this is the best location at the moment

jmacelroy (approved):

PR #393, opened 2023-05-10
docs: update ingress-controller docs for dual

jpf-okteto:
This PR updates ingress-controller docs with latest changes to
ingress-controller architecture introduced at chart 1.8, which includes
dual ingress-controller deployment.

### DIFF
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -1298,13 +1298,18 @@ By default, the Private Endpoint component will inherit the configuration set at
 
 ## Dependencies
 
-Okteto will automatically install [NGINX Ingress Controller](https://kubernetes.github.io/ingress-nginx/) as part of the default installation, using its official [Helm chart](https://kubernetes.github.io/ingress-nginx).
+Okteto will automatically install two instances of [NGINX Ingress Controller](https://kubernetes.github.io/ingress-nginx/) as part of the default installation, using its official [Helm chart](https://kubernetes.github.io/ingress-nginx).
+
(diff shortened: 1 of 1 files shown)


pchico83 (approved):

teresaromero (approved):

ifbyol (commented):
Should we mention the changes in the default-backend? Now, the settings under `ingress-nginx` shouldn't be use. Also, we should probably include in the [`Okteto components` section](https://www.okteto.com/docs/self-hosted/administration/configuration/#okteto-components) the settings we allow for the default backend configuration

@pchico83  WDYT?

maroshii (approved):

pchico83:
> Should we mention the changes in the default-backend? Now, the settings under `ingress-nginx` shouldn't be use. Also, we should probably include in the [`Okteto components` section](https://www.okteto.com/docs/self-hosted/administration/configuration/#okteto-components) the settings we allow for the default backend configuration
> 
> @pchico83 WDYT?

@ifbyol makes sense, but I wouldn't block the release on it. We can create an issue to handle it

ifbyol:
> > Should we mention the changes in the default-backend? Now, the settings under `ingress-nginx` shouldn't be use. Also, we should probably include in the [`Okteto components` section](https://www.okteto.com/docs/self-hosted/administration/configuration/#okteto-components) the settings we allow for the default backend configuration
> > @pchico83 WDYT?
> 
> @ifbyol makes sense, but I wouldn't block the release on it. We can create an issue to handle it

It was done in other PR: https://github.com/okteto/docs/pull/395. Thanks @jpf-okteto !

PR #399, opened 2023-05-12
Added a paragraph to recommend users to set explicitly the setting buildkit.persistence.enabled: true before upgrading to 1.7

ifbyol:
Added a paragraph in the `Before Upgrade` section to recommend to set explicitly the persistence for BuildKit before upgrading to `1.7`. Updated also `1.7` and `1.8` versions

### DIFF
--- a/src/content/self-hosted/install/upgrade.mdx
+++ b/src/content/self-hosted/install/upgrade.mdx
@@ -25,7 +25,15 @@ Please review [the release notes](self-hosted/install/releases.mdx) before upgra
 
 ## Upgrading to Okteto 1.7.x
 
-As part of [1.7.0](self-hosted/install/releases.mdx#170), we moved the configuration settings needed to configure the external storage for the registry. Before, it was configured under `cloud` key and now it has to be configured under `registry` key.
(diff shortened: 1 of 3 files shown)


pchico83 (approved):

src/content/self-hosted/install/upgrade.mdx
```diff
@@ -61,6 +61,13 @@ registry:
 >
 > If you had a custom value for `cloud.secret`, you need to specify the right values under `registry.secret` key.
 
+Also, we have changed the default behavior of BuildKit to don't use a persistent volume. If your BuildKit instance was already using a PVC, it should be kept but we recommend you to include the following key in the configuration file:
```
rberrelleza, line 64: this guidance is odd. it sounds like we don't know if our feature works or not. if we are not confident the volume will stay, then let's add this to the upgrade guide as a must have.

pchico83: +1 to add it to the upgrade guide. We prefer that our users set that property explicitly moving forward

ifbyol: Changed to be the first thing commented in the upgrade to `1.7.x`

PR #400, opened 2023-05-12
Simplify the default installation guide to remove the dependency on a…

pchico83:
… subdomain

Apply the changes from the lego approach initiative to simplify the default installation guide

### DIFF
--- a/src/content/self-hosted/install/deployment.mdx
+++ b/src/content/self-hosted/install/deployment.mdx
@@ -10,7 +10,7 @@ import TabItem from '@theme/TabItem';
 
 ## Configuration file
 
-Before running `helm install`, we recommend that you create a yaml configuration file with your choices about how to install Okteto. This guide will walk you through the most common options. A complete list of configuration options is [available here](self-hosted/administration/configuration.mdx).
+Before running `helm install`, we recommend that you create a yaml configuration file with your choices about how to install Okteto. This guide will walk you through the default installation. A complete list of configuration options is [available here](self-hosted/administration/configuration.mdx).
 
 You can use [this sample configuration file](https://www.okteto.com/docs/self-hosted/install/config.yaml) as a starting point. The different configuration settings are explained below.
 
@@ -27,15 +27,6 @@ Run the following command to obtain your cluster's API server endpoint:
 
 ```
 kubectl config view --minify | grep server
-```
-
-### Domain
-
-You'll need sufficient access to a [subdomain](self-hosted/administration/configuration.mdx#subdomain) to add a wildcard DNS record, such as dev.example.com.
-By default, all endpoints created by Okteto for your development environments will be exposed on the wildcard subdomain you choose.
-
-```yaml
-subdomain: "dev.example.com"
 ```
 
 ### License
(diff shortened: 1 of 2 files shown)


src/content/self-hosted/install/deployment.mdx
```diff
@@ -63,31 +47,33 @@ $ helm install okteto okteto/okteto -f config.yaml --namespace=okteto --create-n
...
+
+Follow these steps to configure the Okteto CLI:
+
+1. Install the Okteto CLI:
+
+   $ curl https://get.okteto.com -sSfL | sh
+
+2. Configure the Okteto CLI context:
+
+   $ okteto context use https://okteto.localtest.me:8443 --token ******
```
rberrelleza, line 68: Where is this domain registered? Do we own it?

rberrelleza (commented):
Is this change consistent with the messaging in the free trial emails and get started guides? (Eg will the application endpoints of the movies app work?)

rberrelleza:
What's the strategy behind not having the license at the beginning? This will make it impossible for us to know who the cluster is.

Could we make the flow:
1 get your free trial
2 do the simplest configuration 
3 follow the UI

src/content/self-hosted/install/deployment.mdx
```diff
@@ -63,31 +47,33 @@ $ helm install okteto okteto/okteto -f config.yaml --namespace=okteto --create-n
...
 
-The default installation is all that's required to follow our [getting started](getting-started.mdx) guide.
-Just remember to replace https://cloud.okteto.com by `https://okteto.SUBDOMAIN` and you will be good to go!
\ No newline at end of file
+   https://www.okteto.com/docs/using-dev-envs/
+
+Happy coding!
+```
+
+Once the chart is installed, follow the onscreen instructions to finalize the configuration.
```
rberrelleza, line 77: I think we need documentation for the different steps, so that the admin knows what they need to prep for the future. I don't think we can relay just on the dashboard to convey this information. @omnomagonz what do you think?

omnomagonz: @rberrelleza Agreed.

I would expect to see the first mentions of these installation steps mentioned in `/self-hosted/install/preparation` where they're outlined similar to the UI. For example:

> Before installing Okteto we recommend you gather the following information and have appropriate levels of access:
> 
> 1. A domain you wish to use
> 2. Certificates to use with the Kubernetes API server
> 3. OAuth access to configure this with Okteto
> 4. External storage environment for the Okteto Registry
> 5. Dedicated Load Balancer for Buildkit

We could then write a few sentences for each to explain what they're used for (so the installer can decide if they want to skip the step or not) and provide context for why they're important for Okteto.

Then this page (`/self-hosted/install/deployment`) could have an intro statement "Be sure to follow the steps in `preparation` first", which would likely prepare them for this part of the docs.

If we're in agreement on that I can draft a PR 👍

pchico83:
> Is this change consistent with the messaging in the free trial emails and get started guides? (Eg will the application endpoints of the movies app work?)

Yes, all the endpoints and getting started guides work

src/content/self-hosted/install/deployment.mdx
```diff
@@ -29,25 +32,6 @@ Run the following command to obtain your cluster's API server endpoint:
...
-You'll need sufficient access to a [subdomain](self-hosted/administration/configuration.mdx#subdomain) to add a wildcard DNS record, such as dev.example.com.
-By default, all endpoints created by Okteto for your development environments will be exposed on the wildcard subdomain you choose.
-
-```yaml
-subdomain: "dev.example.com"
-```
-
-### License
-
-Okteto is free for small teams.  You get all the features of [Okteto](self-hosted.mdx) for up to 3 users.
```
rberrelleza, line 43: Let's mention that it's optional, but encourage to get a license by starting the free trial?

pchico83:
@rberrelleza @omnomagonz are we good to go with this one?

omnomagonz:
@pchico83 I think this PR is fine and we can open additional PRs to make any further changes 😄

omnomagonz (approved):

PR #415, opened 2023-05-31
Remove buildkit LoadBalancer config

teresaromero:
We are recommending to use buildkit within the ingress so this PR removes the documented installation and config.yaml properties for buildkit setting as LoadBalancer


pchico83 (approved):
I would also remove:
```
service.sessionAffinity: The [sessionAffinity](https://kubernetes.io/docs/reference/networking/virtual-ips/#session-affinity) setting for the buildkit service. None by default.
```
from https://www.okteto.com/docs/self-hosted/administration/configuration/#buildkit

PR #418, opened 2023-06-02
Document how to expose buildkit behind Kong

pchico83:
Document how to expose buildkit behind Kong using the right annotations

### DIFF
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -1390,47 +1390,7 @@ Okteto sets specific values on the embedded ingress-nginx chart to enable featur
 
 #### Using Your Own Ingress Controller
 
-You can disable the installation of the main `ingress-nginx` with the following snippet:
-
-```yaml
-ingress-nginx:
-  enabled: false
-```
-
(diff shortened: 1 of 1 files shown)


pchico83:
@rberrelleza do you think this is the right place to document this? or would be better a community post?

rberrelleza (approved):

rberrelleza:
> @rberrelleza do you think this is the right place to document this? or would be better a community post?

I don't think this is the right place for this (or for the bring your own ingress). I think we should create a separate section under self-hosted/administration/configuration/advanced/XXX for the things that a) we have received common requests and b) we plan on supporting officially. 

If this is a single customer ask, i'd go with the community for now. if we get enough requests, we can promote it to the official documentation.

pchico83:
> > @rberrelleza do you think this is the right place to document this? or would be better a community post?
> 
> I don't think this is the right place for this (or for "the bring your own ingress" section either). This section is meant to be a reference for the values that we accept in the configuration, not to explain all the possible scenarios, exceptions, and architecture. I think that we should create a separate section under self-hosted/administration/configuration/advanced/XXX for the topics that a) we have received several requests and b) we plan on supporting officially (meaning that if it doesn't work we'll fix it).
> 
> If this is a single customer ask, we should start with a community post for now. if we get enough requests, we can promote it to the official documentation.

@rberrelleza updated to link to a community guide:
https://community.okteto.com/t/how-do-i-configure-my-own-ingress-controller/887

PR #423, opened 2023-06-20
Remove dependency on "cluster.endpoint" to install the okteto chart

pchico83:
Okteto chart 1.10 removes the need to specify `cluster.endpoint` in the Helm values file.
This PR updates the installation guides to remove this dependency

### DIFF
--- a/src/content/self-hosted/digitalocean.mdx
+++ b/src/content/self-hosted/digitalocean.mdx
@@ -79,14 +79,12 @@ For example:
 
 ```yaml
 license: 1234567890ABCD==
+
 subdomain: dev.example.com
-cluster:
-  endpoint: "https://XXXXXX.aks.microsoft.com"
 
 buildkit:
   persistence:
     enabled: true
-
 ```
 
 ### Installing your Okteto instance
--- a/src/content/self-hosted/eks/config.yaml
+++ b/src/content/self-hosted/eks/config.yaml
@@ -1,12 +1,3 @@
 license:
 
 subdomain: 
-
-cluster:
-  endpoint:
-
-ingress-nginx:
-  controller:
-    service:
-      annotations:
-        service.beta.kubernetes.io/aws-load-balancer-type: nlb
--- a/src/content/self-hosted/aks.mdx
+++ b/src/content/self-hosted/aks.mdx
@@ -34,13 +34,6 @@ We recommend the following specs:
 - A pool with at least 3 `Standard D4` nodes
 - 250 GB per Standard SSD Managed Disk
 
-You'll be using the cluster's API server endpoint when configuring Okteto.
-Run the following command to obtain your cluster's API server endpoint:
-
-```
-kubectl config view --minify | grep server
-```
-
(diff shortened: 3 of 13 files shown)


src/content/self-hosted/digitalocean.mdx
```diff
@@ -78,15 +78,13 @@ Download a copy of the [Okteto DigitalOcean configuration file](https://www.okte
 For example:
 
 ```yaml
-license: 1234567890ABCD==
```
rberrelleza, line 81: Why do we remove the value of the license?

src/content/self-hosted/eks/config.yaml
```diff
@@ -1,12 +1,3 @@
 license:
 
 subdomain: 
-
-cluster:
-  endpoint:
-
-ingress-nginx:
```
rberrelleza, line 8: Why are we removing this part? I don't get how this is related to the cluster endpoint

src/content/self-hosted/digitalocean.mdx
pchico83: That's a fake value, it's better to leave it empty, no?

src/content/self-hosted/eks/config.yaml
pchico83: We shouldn't have this snippet here, the guide isn't using it. At some point we should default to NLB, but there is an open issue about it

src/content/self-hosted/digitalocean.mdx
rberrelleza: no, since we have example values for all the other fields in the values.yaml, we should leave it. This is equally fake than using 'example.com' as the domain

pchico83: @rberrelleza updated

ifbyol:
@pchico83 we will have to wait until `1.11` for this. Even if `1.10` contains the main changes, it's needed a change in the CLI which will come in `2.17.x`. As that CLI release was post-poned, `1.10` won't include it as dependency in the installer, so It won't work properly. `1.11` should include everything we need to make this scenario work properly in all the cases

pchico83:
> @pchico83 we will have to wait until `1.11` for this. Even if `1.10` contains the main changes, it's needed a change in the CLI which will come in `2.17.x`. As that CLI release was post-poned, `1.10` won't include it as dependency in the installer, so It won't work properly. `1.11` should include everything we need to make this scenario work properly in all the cases

@ifbyol thanks for the update, we won't merge this PR until 1.10 is shipped

rberrelleza:
Can we close this PR and then reopen it once we are working on the 1.11 docs? It's easy for someone to make a mistake and merge it since it's approved an everything. Or maybe make it a draft?

pchico83:
> Can we close this PR and then reopen it once we are working on the 1.11 docs? It's easy for someone to make a mistake and merge it since it's approved an everything. Or maybe make it a draft?

@rberrelleza this can be merge now

rberrelleza (approved):

PR #426, opened 2023-06-26
remove docs for versions older than 1.4

rberrelleza:
We don't have any users on versions older than 1.4, so removing the docs


rberrelleza:
@huguestennier is there a way to redirect all the old versions to the latest? (to prevent broken links or old references)

huguestennier:
> @huguestennier is there a way to redirect all the old versions to the latest? (to prevent broken links or old references). Maybe by keeping https://github.com/okteto/docs/pull/426/files#diff-ab8f79b68b7adff7a07db953bf453f3c5aa6ade98d2b1b67d8432b36392489edR24 ?

Yup exactly!

src/content/self-hosted/install/releases.mdx
```diff
@@ -328,229 +328,4 @@ January 19, 2023
 - Fixed blank screen that sometimes appeared when clicking the "Stop development" button
 - Fixed race condition when trying to deploy a development environment that was removed from the catalog
 - Ingresses that are not managed for Okteto won't be deleted when a namespace is put to sleep
-- Fixed how host aliases are injected in the installer pods as the connection to Okteto's API and registry was not using the internal IPs in some scenarios
```
omnomagonz, line 331: @rberrelleza Would it make sense to keep these older release notes for posterity? I don't feel strongly that we need to keep them, but it could be helpful for context gathering for new hires.

Maybe at a minimum we could move these to Notion?

omnomagonz (approved):

PR #427, opened 2023-06-27
Added ?origin=github to GitHub App auth callback 

ifbyol:
Adding that query parameter to the GitHub App auth callback makes that we display the following screen instead of a blank page when users end configuring the repositories accessible by the GitHub App

Captura de pantalla 2023-06-27 a las 9 59 16


pchico83:
@ifbyol are we doing this in our SaaS environments?

rberrelleza (approved):

ifbyol:
> > @ifbyol are we doing this in our SaaS environments?
> 
> @pchico83 I'm not sure, I have pending to review our internal guides when the GitHub App is created to see if we are including it or not

@pchico83 Double checked. Our internal guide already takes this into account when creating the GItHub Apps for SaaS

PR #431, opened 2023-06-30
changing name of key for external storage configuration docs

RinkiyaKeDad:
The secret name should be `key` because that's what is expected by the pods. cc @rberrelleza

### DIFF
--- a/src/content/self-hosted/administration/configuration.mdx
+++ b/src/content/self-hosted/administration/configuration.mdx
@@ -588,7 +588,7 @@ Create a secret with the value of `STORAGE_ACCOUNT_ACCESS_KEY` using the command
 
 
 ```console
-$ kubectl create secret generic okteto-registry-secret --namespace=okteto --from-literal=key=STORAGE_ACCOUNT_ACCESS_KEY
+$ kubectl create secret generic okteto-registry-secret --namespace=okteto --from-literal=secretKey=STORAGE_ACCOUNT_ACCESS_KEY
 ```
 
 Then, add the following configuration to your Okteto helm values file:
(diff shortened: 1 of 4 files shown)


ifbyol:
@RinkiyaKeDad The value you have to specify there depends on the value configured here: https://www.okteto.com/docs/self-hosted/administration/configuration/#registry under the key secret on `accessKey` and `secretKey`.

By default, our chart value uses `secretKey`, so not sure about changing it

RinkiyaKeDad:
image

@ifbyol if we use `secretKey` then we were getting this error. You can see in the message that it expects the name of the file where the secret is stored to be `key` and that will only be possible if we do `--from-literal=key=STORAGE_ACCOUNT_ACCESS_KEY` instead of `--from-literal=secretKey=STORAGE_ACCOUNT_ACCESS_KEY` no?

ifbyol:
> image
> 
> @ifbyol if we use `secretKey` then we were getting this error. You can see in the message that it expects the name of the file where the secret is stored to be `key` and that will only be possible if we do `--from-literal=key=STORAGE_ACCOUNT_ACCESS_KEY` instead of `--from-literal=secretKey=STORAGE_ACCOUNT_ACCESS_KEY` no?

Yes, for GKE it's using directly `key` instead of the field configured for in the section I mentioned above. For the rest of providers we need to keep `accessKey` but for GKE we need to change it to `key`

RinkiyaKeDad:
@ifbyol I updated the PR to just change it to `key` for GCP, PTAL.

ifbyol (approved):
Thanks Arsh! Looks good!

omnomagonz (approved):

PR #432, opened 2023-07-03
Update CLI reference for 2.17.0

teresaromero:
Remaining changes for releaes 2.17.0

### DIFF
--- a/src/content/reference/cli.mdx
+++ b/src/content/reference/cli.mdx
@@ -383,7 +383,7 @@ Use the arrow keys to navigate: ↓ ↑ → ←
 
 #### create
 
-Create a namespace. By default the command will switch to the new namespace, add `--use=false` for just creating the namespace and keeping the current one active.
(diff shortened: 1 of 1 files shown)


AdrianPedriza (approved):

jLopezbarb (approved):

src/content/reference/cli.mdx
```diff
@@ -383,7 +383,7 @@ Use the arrow keys to navigate: ↓ ↑ → ←
 
 #### create
 
-Create a namespace.
+Create a namespace. By default the command will switch to the new namespace, add `--use=false` for just creating the namespace and keeping the current one active.
```
omnomagonz, line 386:
```suggestion
Create a namespace. By default the command will switch to the new namespace, add `--use=false` to create the namespace, and it will keep the current namespace active.
```

Non-blocking suggestion based on what I think I understand of this copy 😄

PR #443, opened 2023-07-07
New supported flag on okteto preview destroy

teresaromero:
Docs for https://github.com/okteto/okteto/pull/3776

### DIFF
--- a/src/content/reference/cli.mdx
+++ b/src/content/reference/cli.mdx
@@ -572,6 +572,10 @@ You need to be [logged in](reference/cli.mdx#context) to Okteto before running t
 Only users with administration privileges can destroy a global preview environment.
 If the manifest includes `destroy` commands, they will be executed as part of this command.
 
+| Options                                   |    Type    | Description                                                                                                                                                                                                                                               |
+| :---------------------------------------- | :--------: | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| _--wait_                                  |   (bool)   | Wait until the preview deletion finishes (defaults to false)                                                                                                                                                                                            |
+
 #### endpoints
 
 List the endpoints of a preview environment.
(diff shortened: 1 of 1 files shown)


src/content/reference/cli.mdx
```diff
@@ -572,6 +572,10 @@ You need to be [logged in](reference/cli.mdx#context) to Okteto before running t
 Only users with administration privileges can destroy a global preview environment.
 If the manifest includes `destroy` commands, they will be executed as part of this command.
 
+| Options                                   |    Type    | Description                                                                                                                                                                                                                                               |
+| :---------------------------------------- | :--------: | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| _--wait_                                  |   (bool)   | Wait until the preview deletion finishes (defaults to false)                                                                                                                                                                                            |
```
nitishfy, line 577:
```suggestion
| _--wait_                                  |   (bool)   | Wait until the preview deletion finishes (defaults to true)                                                                                                                                                                                            |
```

The default value is set to true to avoid breaking changes.

nitishfy (changes requested):

pchico83 (approved):

src/content/reference/cli.mdx
```diff
@@ -572,6 +572,10 @@ You need to be [logged in](reference/cli.mdx#context) to Okteto before running t
 Only users with administration privileges can destroy a global preview environment.
 If the manifest includes `destroy` commands, they will be executed as part of this command.
 
+| Options                                   |    Type    | Description                                                                                                                                                                                                                                               |
+| :---------------------------------------- | :--------: | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| _--wait_                                  |   (bool)   | Wait until the preview deletion finishes (defaults to true)                                                                                                                                                                                            |
```
rberrelleza, line 577: I think we should remove that part since the default is to not wait if the flag is not available right?

```suggestion
| _--wait_                                  |   (bool)   | Wait until the preview deletion finishes                                                                                                                                                                                  |
```

andreafalzetti: @rberrelleza by default it will wait, because otherwise it would be a breaing change since currently it wait, with 3.0 we'll set default to false

rberrelleza: then how do you tell it not to wait? `--wait=false` ?

nitishfy: Yes, `--wait=false` will schedule the destroy.

rberrelleza: Ah thanks for the explanation!

rberrelleza (approved):

rberrelleza (approved):

Issue #408, opened 2023-05-23
Add docs about hybrid development


AdrianPedriza:
we are introducing hybrid development feature. This feature requires update manifest docs.

epic -> https://www.notion.so/okteto/Okteto-CLI-3-0-Epic-1-Hybrid-Development-a3dcb3a6aa16411e91cde48f6a187e24?pvs=4

PR #447, opened 2023-07-10
add docs about dev mode field in manifest

AdrianPedriza:
Fixes: https://github.com/okteto/docs/issues/408

### DIFF
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -511,13 +511,24 @@ metadata:
 
 #### mode (string, optional)
 
-Defines how `okteto up` will behave.
+Defines how `okteto up` will setup the development container and behave within the cluster.
 
 Currently two options are available:
-- `sync`: (default) Run service in a remote development container. Synchronizes your code with the container.
-- `hybrid`: Runs service in your local computer instead of in a remote development container. Traffic needs to be "reversed" in order to communicate with the rest of the components running in your cluster.
+- `sync`: (default) Sync & run your code in a remote development container.
+- `hybrid`: Run your code on your local computer instead of in a remote development container, while rest of your service components runs in your remote cluster.
 
-The motivation for this field comes from the fact that adopting a `hybrid` approach can sometimes provide a faster development experience because in general the local machine is faster than our development containers, saving e.g. service build time. On the other hand, it should be noted that developing with the service running locally breaks replicability between OS in addition to the added latency when accessing the rest of the services in the cluster. 
+Most use-cases should use default `sync` mode, but `hybrid` may be desired to achieve a faster development experience in some scenarios:
+
+|   | `sync` | `hybrid` |
+|---|---|---|
+| Soure code  | Uses Syncthing to sync files bidirectionally with your remote development container | Uses local source code directly, no files syncing needed |
+| `command` context | Runs in the _remote development container_ | Runs on your _local machine_, so traffic need to be rerouted from/to cluster, e.g. using `reverse` option |
+
+##### `hybrid` motivation and example
+
+A typical use-case and motivation for using `hybrid` is when developing a service that require "very fast" file-change-to-reload times. Frontend development tooling (e.g. Webpack) is a common example. Since hybrid mode avoids having to sync file changes to remote development container and wait for the container to detect file changes to trigger a rebuild, it's typically order-of-magnitude faster than `sync` mode.
+
+On the other hand, with the service running locally breaks replicability between OS, and may add latency when accessing the rest of the services in the cluster.  But it's usually an OK tradeoff for development purposes.
 
 If you want to use the hybrid mode, you only need to configure the necessary ports to expose the service running on your local machine to the development container in remote so that the rest of the components running on the cluster can communicate with the development container as if it were the one running the application remotely. To do that, you can make use of the [reverse](reference/manifest.mdx#reverse) field, which will expose the indicated port of your local machine to the desired port on the development container.
 
@@ -528,7 +539,7 @@ dev:
     reverse: 
     - 3000:8080
     workdir: ./frontend
-    command: ["yarn start"]
+    command: ["npx webpack watch --mode development"]
 ```
 
 in the example shown we have indicated that we develop the `frontend` service locally using the `mode: hybrid` field.  The service running locally will execute the command defined in the `command` field in the directory indicated in `workdir`. In addition, the application will expose on port `3000` of the development container what it exposes on port `8080` of the local machine, making it possible for other components running on the cluster to access the service as if it were running remotely.
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -497,7 +497,7 @@ lifecycle:
 - `postStart`: specifies if the postStart event is enabled when running `okteto up` (default: `false`).
 - `postStop`: specifies if postStop event is enabled when running `okteto up` (default: `false`).
 
-#### metadatassss (object, optional)
+#### metadata (object, optional)
 
 The metadata field allows to inject labels and annotations into your development container.
 
--- a/src/content/reference/manifest.mdx
+++ b/src/content/reference/manifest.mdx
@@ -497,7 +497,7 @@ lifecycle:
 - `postStart`: specifies if the postStart event is enabled when running `okteto up` (default: `false`).
 - `postStop`: specifies if postStop event is enabled when running `okteto up` (default: `false`).
 
-#### metadata (object, optional)
+#### metadatassss (object, optional)
 
 The metadata field allows to inject labels and annotations into your development container.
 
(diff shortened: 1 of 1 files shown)


src/content/reference/manifest.mdx
```diff
@@ -509,6 +509,41 @@ metadata:
     custom.label/dev: "true"
 ```
 
+#### mode (string, optional)
+
+Defines how `okteto up` will setup the development container and behave within the cluster.
+
+Currently two options are available:
+- `sync`: (default) Sync & run your code in a remote development container.
+- `hybrid`: Run your code on your local computer instead of in a remote development container, while rest of your service components runs in your remote cluster.
```
omnomagonz, line 518:
```suggestion
- `hybrid`: Run your code on your local computer instead of in a remote development container, while rest of your service components runs in your remote cluster.
```

"Run your code on your local computer" -- isn't this how Okteto works today? This makes it sound like if using `sync` (default) then your code runs in the remote dev container. Just want to make sure I understand correctly before considering a suggested change here 😄

src/content/reference/manifest.mdx
```diff
@@ -509,6 +509,41 @@ metadata:
...
+
+A typical use-case and motivation for using `hybrid` is when developing a service that require "very fast" file-change-to-reload times. Frontend development tooling (e.g. Webpack) is a common example. Since hybrid mode avoids having to sync file changes to remote development container and wait for the container to detect file changes to trigger a rebuild, it's typically order-of-magnitude faster than `sync` mode.
+
+On the other hand, with the service running locally breaks replicability between OS, and may add latency when accessing the rest of the services in the cluster.  But it's usually an OK tradeoff for development purposes.
+
+If you want to use the hybrid mode, you only need to configure the necessary ports to expose the service running on your local machine to the development container in remote so that the rest of the components running on the cluster can communicate with the development container as if it were the one running the application remotely. To do that, you can make use of the [reverse](reference/manifest.mdx#reverse) field, which will expose the indicated port of your local machine to the desired port on the development container.
```
omnomagonz, line 533:
```suggestion
To use `hybrid` mode, you need to configure the necessary ports to expose the service running on your local machine to the remote development container. This ensures all components running on the cluster can communicate with the development container and your local services by extension. 

To do this you can make use of the [reverse](reference/manifest.mdx#reverse) field, which will expose the indicated port of your local machine to the desired port on the development container.
```

src/content/reference/manifest.mdx
rberrelleza: sync is how okteto up behaves today. 'hybrid' is the new model where the process runs in your local machine, but the rest of the application runs on the cluster.

src/content/reference/manifest.mdx
```diff
@@ -509,6 +509,41 @@ metadata:
...
+- `hybrid`: Run your code on your local computer instead of in a remote development container, while rest of your service components runs in your remote cluster.
+
+Most use-cases should use default `sync` mode, but `hybrid` may be desired to achieve a faster development experience in some scenarios:
+
+|   | `sync` | `hybrid` |
+|---|---|---|
+| Soure code  | Uses Syncthing to sync files bidirectionally with your remote development container | Uses local source code directly, no files syncing needed |
+| `command` context | Runs in the _remote development container_ | Runs on your _local machine_, so traffic need to be rerouted from/to cluster, e.g. using `reverse` option |
+
+##### `hybrid` motivation and example
```
rberrelleza, line 527: @omnomagonz I don't think this kind of docs should "live" here. This makes it harder to navigate the reference. Maybe we need to have a separate page explaining the motivation behind 'okteto up', and then talk in-depth about the modes and when to choose one vs the other?

src/content/reference/manifest.mdx
```diff
@@ -509,6 +509,41 @@ metadata:
...
 
+#### mode (string, optional)
+
+Defines how `okteto up` will setup the development container and behave within the cluster.
+
+Currently two options are available:
+- `sync`: (default) Sync & run your code in a remote development container.
+- `hybrid`: Run your code on your local computer instead of in a remote development container, while rest of your service components runs in your remote cluster.
+
+Most use-cases should use default `sync` mode, but `hybrid` may be desired to achieve a faster development experience in some scenarios:
```
rberrelleza, line 520: This needs to be more specific. Can we talk more concretely about why pick one vs the other? If not, I think it's better to remove this and describe the options without having an opinion on which one to use.

src/content/reference/manifest.mdx
```diff
@@ -509,6 +509,41 @@ metadata:
...
+
+|   | `sync` | `hybrid` |
+|---|---|---|
+| Soure code  | Uses Syncthing to sync files bidirectionally with your remote development container | Uses local source code directly, no files syncing needed |
```
rberrelleza, line 524: both sync and hybrid "Uses local source code directly".  I don't think this is a good way to explain the difference. The main difference is that on sync, the code is synchronized and development runs in a container, while on hybrid, the service under development runs in the local computer, and it "connects" with the rest of the application in the cluster, right?

src/content/reference/manifest.mdx
```diff
@@ -509,6 +509,41 @@ metadata:
...
+| `command` context | Runs in the _remote development container_ | Runs on your _local machine_, so traffic need to be rerouted from/to cluster, e.g. using `reverse` option |
+
+##### `hybrid` motivation and example
+
```
rberrelleza, line 528: In which scenario you don't want a fast file-change-to-reload? Speed is not the right way to explain the reason. No one wants to pick a "slower" action.

src/content/reference/manifest.mdx
```diff
@@ -509,6 +509,41 @@ metadata:
...
+A typical use-case and motivation for using `hybrid` is when developing a service that require "very fast" file-change-to-reload times. Frontend development tooling (e.g. Webpack) is a common example. Since hybrid mode avoids having to sync file changes to remote development container and wait for the container to detect file changes to trigger a rebuild, it's typically order-of-magnitude faster than `sync` mode.
```
rberrelleza, line 529: Do we have data to back this up? If I read this, my first reaction would be never using "sync" mode. Why would I pick something 10 times slower?

src/content/reference/manifest.mdx
AdrianPedriza: yes, that's it. Maybe is a little bit confusing this table yeah. I don't think it add too much value

src/content/reference/manifest.mdx
AdrianPedriza: should we talk about it on the dedicated page/blog?

AdrianPedriza:
In general I also think that except for describing the available values in each of the fields the rest should go on another page. Maybe adding a link to a blog using the hybrid mode and explaining why it can be beneficial in some cases? cc @omnomagonz @rberrelleza @RinkiyaKeDad

src/content/reference/manifest.mdx
RinkiyaKeDad: I agree! I'm open to having this in a blog, but then those won't be easy to discover for someone trying to learn more about the feature. Maybe a separate doc page like we did for Catalog. But then that begs the question where do we draw the line on what gets a separate page and what doesn't.

src/content/reference/manifest.mdx
RinkiyaKeDad: I agree with @rberrelleza that we should remove it here if we don't explain it in detail. More thoughts here: https://github.com/okteto/docs/pull/447/files#r1259847016

RinkiyaKeDad:
I'm open to having a blog for this! It could be a nice idea to get visibility for this feature. But my concern is we still need something in the docs which explains this. For example, someone looking to learn more about this feature when implementing might not stumble on the blog, but they will definitely go to the docs to search about this. 

Now we could link the blog from the docs, but I haven't seen a lot of technical docs do that and I think that's because the target audience and the tone for blogs is usually different from docs. If we feel this feature is relevant enough, maybe we could have a separate page in the docs?

omnomagonz:
I opened https://github.com/okteto/docs/pull/450 since I thought it might be cleaner than altering this one.

Please let me know if anyone prefers to consolidate into this PR and/or feel free to comment on the new PR

src/content/reference/manifest.mdx
viddo: I think this is resolved? Feel free to reopen it again and clarify suggested change. :)

viddo:
Thanks for the input everyone! Based on additional offline feedback, I've reduced the scope to just be about the reference manifest. 

I've resolved all the open but now outdated comment threads. Let's continue to iterate & reify on all that separately then.

Please have a review again and let me know if you you think are any other blockers.

viddo:
Closing this one in favor of #450 then, let's continue there.
Sorry for any noisy notifications (deleted prev comment since I had a cached outdated state of this PR…).

PR #450, opened 2023-07-17
Add hybrid mode content

omnomagonz:
Based on https://github.com/okteto/docs/pull/447 I'm proposing we add the `hybrid` mode content in these areas and using this language.

Opening a separate PR seemed cleaner 😄

### DIFF
--- a/src/content/cloud/okteto-cli.mdx
+++ b/src/content/cloud/okteto-cli.mdx
@@ -5,6 +5,9 @@ sidebar_label: Okteto CLI
 id: okteto-cli
 ---
 
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
 [Dev Environments](reference/development-environments.mdx) allow you to develop your deployed application locally. You continue coding in your favorite IDE and see the changes live on the deployed version of your application as soon as you hit save.
 Okteto CLI is an [open source](https://github.com/okteto/okteto) client-side only tool that allows you to launch Dev Environments in any Kubernetes cluster: your own, [Okteto Cloud](index.md), or [Okteto Self-Hosted](self-hosted.mdx).
 
@@ -28,6 +31,57 @@ This command has two modes: `sync` (default mode) and `hybrid` which are detaile
 | Source code       | Source code is edited locally and synchronizes bidirectionally (via Syncthing) with a copy of the source code in the remote development container. | Source code is edited locally for a chosen set of microservices. |
 | App location      | The application exists entirely within the remote development container. | Your application is split between your local environment and the remote development container based on your configuration. |
 | `command` context | CLI commands run in the remote development container. | CLI commands run on your local machine, so traffic need to be rerouted to/from the cluster, e.g. using the reverse option |
+
+Below is a reference to help you configure your [`dev` section](/reference/manifest/#dev-object-optional) based on the mode you select. **Standard** configuration options work in both modes and **Sync** options work *only* when `sync` is selected.
+
+For example, if you select `hyrid` mode, you can only use the configuration options in your manifest `dev` section listed under **Standard**. Any configuration options in the **Sync** list will *not work*.
+
+<details>
+    <summary>Configuration reference: dev section</summary>
+        <Tabs>
+          <TabItem value="Standard" label="Standard" default>
+            <ul>
+                <li>annotations</li>
+                <li>autocreate</li>
+                <li>command</li>
+                <li>container</li>
+                <li>envFiles</li>
+                <li>environment</li>
+                <li>environment</li>
+                <li>forward</li>
+                <li>metadata</li>
+                <li>mode</li>
+                <li>reverse</li>
+                <li>selector</li>
+                <li>workdir</li>
+            </ul>
+          </TabItem>
+          <TabItem value="Sync" label="Sync">
+            <ul>
+                <li>affinity</li>
+                <li>context</li>
+                <li>externalVolumes</li>
+                <li>image</li>
+                <li>imagePullPolicy</li>
+                <li>initContainer</li>
+                <li>initFromImage</li>
+                <li>lifecycle</li>
+                <li>namespace</li>
+                <li>nodeSelector</li>
+                <li>persistentVolume</li>
+                <li>push</li>
+                <li>replicas</li>
+                <li>secrets</li>
+                <li>securityContext</li>
+                <li>serviceAccount</li>
+                <li>sshServerPort</li>
+                <li>sync</li>
+                <li>tolerations</li>
+                <li>volumes</li>
+            </ul>
+          </TabItem>
+        </Tabs>
+</details>
 
 ### When to use `hybrid` development mode
 
--- a/src/content/cloud/okteto-cli.mdx
+++ b/src/content/cloud/okteto-cli.mdx
@@ -17,6 +17,46 @@ Learn more about the Okteto CLI
 - [FAQs](reference/faqs.mdx)
 - [Known Issues](reference/known-issues.mdx)
 
+## okteto up
+
+The `okteto up` command is one of Okteto CLI's main features and key to the Okteto development experience. This command uses the [Okteto Manifest](/reference/manifest) you define to build, deploy, and synchronize your local files with the remote development container.
+
+This command has two modes: `sync` (default mode) and `hybrid` which are detailed below.
+
+|                   | `sync` (default) | `hybrid` |
+| :---------------- | :--------------- | :------- |
+| Source code       | Source code is edited locally and synchronizes bidirectionally (via Syncthing) with a copy of the source code in the remote development container. | Source code is edited locally for a chosen set of microservices. |
+| App location      | The application exists entirely within the remote development container. | Your application is split between your local environment and the remote development container based on your configuration. |
+| `command` context | CLI commands run in the remote development container. | CLI commands run on your local machine, so traffic need to be rerouted to/from the cluster, e.g. using the reverse option |
+
+### When to use `hybrid` development mode
+
+A typical use case for using `hybrid` development mode is when the filesync or rebuild is significantly slower on the remote dev container, such as when using Webpack where this becomes very noticeable compared to your local environment. 
+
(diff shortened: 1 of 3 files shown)


src/content/reference/manifest.mdx
```diff
@@ -509,6 +509,31 @@ metadata:
...
+```yaml
+dev:
+  frontend:
+    mode: hybrid
+    reverse:
+    - 3000:8080
+    workdir: ./frontend
+    command: ["npx webpack watch --mode development"]
+```
+
```
rlamana, line 536: With the new `hybrid` not all parameters in the `dev` section apply in this mode. Shouldn't we add a clarification in every one of these parameters stating if it's supported in `sync` and/or `hybrid` mode?

Example:

```
### envFiles
> Supported in both `sync` and `hybrid` modes.
...

### affinity
> Only supported in `sync` mode.
...
```

omnomagonz: @rlamana Ah yes. I think that's a good and important clarification to add.

Are those the only two exceptions or do we need to compile a complete list?

rlamana: Not, we have to take into account all of them. The list is compiled in the spec doc here: https://www.notion.so/okteto/Okteto-CLI-3-0-7fbd3a6a566145d3a4930d3c15fc642d?pvs=4#37aeeb1c05824203ac5f8132caf76e1f

@andreafalzetti @teresaromero you might have more context if that table there is updated after the work done there or is there anything we need to change?

andreafalzetti: @omnomagonz @rlamana 

These are the fields that are supported in `sync` but not in `hybrid`:

- affinity
- context
- externalVolumes
- image
- imagePullPolicy
- initContainer
- initFromImage
- lifecycle
- namespace
- nodeSelector
- persistentVolume
- push
- replicas
- secrets
- securityContext
- serviceAccount
- sshServerPort
- sync
- tolerations
- volumes


These fields work in both modes:

- annotations
- autocreate
- command
- container
- envFiles
- environment
- environment
- forward
- metadata
- mode
- reverse
- selector
- workdir

These fields work in `sync` but we're not sure if they make sense for `hybrid`:
- interface
- probes
- remote
- timeout

This field exists in code but not documented and not sure if it should be supported:
- args

Should work in both but deprecated:
- healthchecks
- labels

rlamana: Thanks a lot @andreafalzetti for the updated list. We will need to clarify those that we are not sure:
- interface
- probes
- remote
- timeout
As discussed offline maybe @pchico83 can help us with more context on these, and soon before he leaves on PTO.

omnomagonz: Thank you both for this context 🙇‍♂️ I'll commit a change with how I propose we represent this and let's continue iterating 🙂

omnomagonz: For clarity:

> These fields work in sync but we're not sure if they make sense for hybrid:

Does this mean those options work in both but only are compatible/make sense for `sync`? So a user could use these options in `hybrid` mode but we're not sure how they'll work or how they'll benefit the user?

omnomagonz: Resolving this conversation; please see latest commit: https://github.com/okteto/docs/pull/450/commits/d45fc31ba9222085aca74aa816e2e4cdbbe00831 and add additional comments :)

I left out some of the nuanced details like "deprecated but works in both" and "exists in code but not documented". We can iterate to add more clarity on these but I didn't want to get blocked on nuance for now

src/content/reference/manifest.mdx
```diff
@@ -509,6 +509,31 @@ metadata:
     custom.label/dev: "true"
 ```
 
+## mode (string, optional)
```
AdrianPedriza, line 512:
```suggestion
#### mode (string, optional)
```
needs to be formatted under the dev section instead of in the root of the content tree

src/content/reference/manifest.mdx
```diff
@@ -509,6 +509,31 @@ metadata:
...
+
+Defines how `okteto up` will setup the development container and behave within the cluster.
+
+Currently two options are available:
+
+- `sync`: (default) Sync & run your code in a remote development container.
+- `hybrid`: Run your code on your local computer instead of in a remote development container, while rest of your service components runs in your remote cluster.
+
+|                   | `sync` (default) | `hybrid` |
```
AdrianPedriza, line 521: I think we have the same table in `cloud/okteto-cli`. Should we link that table here instead of repeating it?

src/content/reference/cli.mdx
```diff
@@ -709,8 +709,9 @@ The `status` command should be run from the same location than `okteto up`.
 
 ### up
 
-Activate a development container.
-If needed, `okteto up` builds the images and runs the deploy commands defined in your [Okteto Manifest](reference/manifest.mdx).
+The `okteto up` command is one of Okteto’s key features. This command uses the [Okteto Manifest](/reference/manifest) you define to build, deploy, and synchronize your local files with the remote development container.
+
+This command has two modes: `sync` (default mode) and `hybrid` which are detailed in the manifest reference: [mode](/reference/manifest#mode-string-optional).
```
AdrianPedriza, line 714: Is the link to `mode` broken? In the preview, I cannot be redirected to that section

src/content/cloud/okteto-cli.mdx
```diff
@@ -17,6 +20,97 @@ Learn more about the Okteto CLI
...
+| :---------------- | :--------------- | :------- |
+| Source code       | Source code is edited locally and synchronizes bidirectionally (via Syncthing) with a copy of the source code in the remote development container. | Source code is edited locally for a chosen set of microservices. |
+| App location      | The application exists entirely within the remote development container. | Your application is split between your local environment and the remote development container based on your configuration. |
+| `command` context | CLI commands run in the remote development container. | CLI commands run on your local machine, so traffic need to be rerouted to/from the cluster, e.g. using the reverse option |
+
+Below is a reference to help you configure your [`dev` section](/reference/manifest/#dev-object-optional) based on the mode you select. **Standard** configuration options work in both modes and **Sync** options work *only* when `sync` is selected.
+
+For example, if you select `hyrid` mode, you can only use the configuration options in your manifest `dev` section listed under **Standard**. Any configuration options in the **Sync** list will *not work*.
```
AdrianPedriza, line 37: IMO it seems strange to list under standard the hybrid mode (considering that it is not the default mode). Shouldn't we refer to the modes here and in the section below as `hybrid` or `sync` under this context to avoid confusion?

rlamana, line 37: typo:
```suggestion
For example, if you select `hybrid` mode, you can only use the configuration options in your manifest `dev` section listed under **Standard**. Any configuration options in the **Sync** list will *not work*.
```

rlamana: I would avoid saying things like "will not work", I'd say "some configuration options do not apply an will be ignore", or something along those line.

The thing is, all that sounds like we don't want to support it or like if it is unfinished, but the reality is that some options don't make sense given the nature of the selected "mode".

On the other hand, there is a use case that a user might want to switch from one mode to another (maybe testing what mode works best for them or just because it is better suited for an specific case). In that case, they don't need to remove all those options that don't apply in their manifest, those not supported will just be ignored.

Hope this help to explain better the nuances of both modes.

rlamana: @omnomagonz I also don't get what exactly "Standard" means?

src/content/reference/manifest.mdx
rlamana: @omnomagonz "not sure if they make sense for hybrid" meant that we are still deciding if the will fall under the "ignored/unsupported" fields or if we will implement it in both modes. But there will be a final resolution for all those so that the final list will only have "supported or not supported" for each mode.

@andreafalzetti what is the current status of those fields?

src/content/reference/manifest.mdx
omnomagonz: My vote would be to repeat it because I think it's important in both sections and is succinct.

If the content was more verbose then I'd say link to it, but since this is fairly concise I think it's worth saving the user a trip to another doc page so they can stay in the same context.

Happy to hear contrary opinion though 😄

src/content/reference/cli.mdx
omnomagonz:
```suggestion
This command has two modes: `sync` (default mode) and `hybrid` which are detailed in the manifest reference: [mode](/reference/manifest/#mode-string-optional).
```
It was broken due to a missing `/` 🤦‍♂️ 

Great catch!

src/content/cloud/okteto-cli.mdx
omnomagonz: The reason I landed on Standard vs. Sync (which I agree doesn't sound right in hindsight) was because it didn't make sense to list: Standard (works for both), sync, and hybrid (duplicate of what's listed in Standard).

One option is to list Sync vs. Hybrid and just list all compatible commands even if it means duplicating some.

And I'll make some adjustments to language to avoid "will not work" 👍

src/content/reference/manifest.mdx
rlamana: In my opinion, besides adding a table I will still add a "hint" on every single parameter. It is common to just go and check a parameter in the docs to learn about it. If the user doesn't have that info there (and assuming they are aware of these two modes and know there is a tablefor it), users will have to go to look for the table and see if the parameter is there, and then maybe go back to the parameter description to continue reading about it. I find the table less important since, you usually go seeking for documentation on a particular item. It is pretty common to see this type of support descriptions in docs of this kind.

andreafalzetti: @rlamana for `remote` I was able to verify that it works for hybrid too. `timeout` for me doesn't work in sync either, so I've created https://github.com/okteto/okteto/issues/3849

still not sure about these tbh:
- interface
- probes

as soon as I can verify I let you guys know.

rlamana: @andreafalzetti So after our daily today, I think `probes` was clarified and it's good to go. What about `interface` then?

omnomagonz:
Closing this PR for now

PR #492, opened 2023-09-14
docs: release notes for 1.12.1

jpf-okteto:


src/content/self-hosted/install/releases.mdx
```diff
@@ -5,6 +5,16 @@ sidebar_label: Release Notes
...
 ---
 
+## 1.12.1
+14 September 2023
+
+This version is compatible with Kubernetes versions 1.24 to 1.26.
+
+### Bugfixes
+
+- Update Kubernetes internal dependencies to fix a possible crash/panic when Okteto lists the cluster API Resources. <!-- 7162, 7166 -->
```
ifbyol, line 15: I think we should comment in which operation it was failing, at least the one we have detected to provide more context to users to know if they faced it or not.

In this case it was during destroy namespace or preview operations

PR #496, opened 2023-09-22
Added configuration settings to configure NLB for Okteto in default installation guide for EKS

ifbyol:
In the configuration file we offer to download directly from the docs (https://github.com/okteto/docs/blob/main/src/content/self-hosted/eks/config.yaml) we include the settings to configure an NLB, but the snippet in the page doesn't have it, so as NLB is preferred, I'm adding also those settings in the snippet in case users copy-paste directly from there instead of downloading the config example file

### DIFF
--- a/src/content/self-hosted/eks.mdx
+++ b/src/content/self-hosted/eks.mdx
@@ -91,6 +91,8 @@ ingress-nginx:
         service.beta.kubernetes.io/aws-load-balancer-type: nlb
 ```
 
+Annotation `service.beta.kubernetes.io/aws-load-balancer-type: nlb` is used to tell AWS to create a [Network Load Balancer (NLB)](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html) instead of the default [Classic Load Balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html) to expose the ingress controller. We recommend to use a NLB as it provides benefits like: websocket support, static IPs, and better performance. You can see a full comparasion ebtween the load balancers [here](https://aws.amazon.com/elasticloadbalancing/features/).
+
 ### Installing your Okteto instance
 
 Install the latest version of Okteto by running:
--- a/src/content/self-hosted/eks.mdx
+++ b/src/content/self-hosted/eks.mdx
@@ -91,7 +91,7 @@ ingress-nginx:
         service.beta.kubernetes.io/aws-load-balancer-type: nlb
 ```
 
-Annotation `service.beta.kubernetes.io/aws-load-balancer-type: nlb` is used to tell AWS to create a [Network Load Balancer (NLB)](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html) instead of the default [Classic Load Balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html) to expose the ingress controller. We recommend to use a NLB as it provides benefits like: websocket support, static IPs, and better performance. You can see a full comparasion between the load balancers [here](https://aws.amazon.com/elasticloadbalancing/features/).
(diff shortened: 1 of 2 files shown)


src/content/self-hosted/eks.mdx
```diff
@@ -83,6 +83,12 @@ license: 1234567890ABCD==
 subdomain: dev.example.com
 cluster:
   endpoint: "https://XXXXXX.gr7.us-west-2.eks.amazonaws.com"
+
+ingress-nginx:
+  controller:
+    service:
+      annotations:
+        service.beta.kubernetes.io/aws-load-balancer-type: nlb
```
pchico83, line 91: Could we add a short explanation about what this annotation does, and why NLB is preferred over ELB?

pchico83 (approved):

pchico83 (approved):

versioned_docs/version-1.12/self-hosted/eks.mdx
```diff
@@ -83,8 +83,16 @@ license: 1234567890ABCD==
...
   endpoint: "https://XXXXXX.gr7.us-west-2.eks.amazonaws.com"
+
+ingress-nginx:
+  controller:
+    service:
+      annotations:
+        service.beta.kubernetes.io/aws-load-balancer-type: nlb
 ```
 
+Annotation `service.beta.kubernetes.io/aws-load-balancer-type: nlb` is used to tell AWS to create a [Network Load Balancer (NLB)](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html) instead of the default [Classic Load Balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html) to expose the ingress controller. We recommend to use a NLB as it provides benefits like: websocket support, static IPs, and better performance. You can see a full comparasion between the load balancers [here](https://aws.amazon.com/elasticloadbalancing/features/).
```
rberrelleza, line 94:
```suggestion
The `service.beta.kubernetes.io/aws-load-balancer-type: nlb`  annotation is used to tell AWS to create a [Network Load Balancer (NLB)](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html) instead of the default [Classic Load Balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html) to expose the ingress controller. We recommend to use a NLB as it provides benefits like: websocket support, static IPs, and better performance. You can see a full comparasion between the load balancers [here](https://aws.amazon.com/elasticloadbalancing/features/).
```

rberrelleza: Is the NLB a recommendation or mandatory? (e.g. will the logs in the uI work if you have a classic LB?)

src/content/self-hosted/eks.mdx
```diff
@@ -83,8 +83,16 @@ license: 1234567890ABCD==
...
 ```
 
+Annotation `service.beta.kubernetes.io/aws-load-balancer-type: nlb` is used to tell AWS to create a [Network Load Balancer (NLB)](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html) instead of the default [Classic Load Balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html) to expose the ingress controller. We recommend to use a NLB as it provides benefits like: websocket support, static IPs, and better performance. You can see a full comparasion between the load balancers [here](https://aws.amazon.com/elasticloadbalancing/features/).
```
rberrelleza, line 94:
```suggestion
Annotation `service.beta.kubernetes.io/aws-load-balancer-type: nlb` is used to tell AWS to create a [Network Load Balancer (NLB)](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html) instead of the default [Classic Load Balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html) to expose the ingress controller. We recommend to use a NLB as it provides benefits like: websocket support, static IPs, and better performance. You can see a full comparasion between the load balancers [here](https://aws.amazon.com/elasticloadbalancing/features/).
```

versioned_docs/version-1.12/self-hosted/eks.mdx
ifbyol: @rberrelleza It is a recommendation. Classic load balancer works with Okteto

omnomagonz (approved):

PR #498, opened 2023-09-22
1.13 versionsed docs

maroshii:
Adds 1.13 versioned docs. Only Algolia reindexing is pending.


rberrelleza (approved):

docusaurus.config.js
```diff
@@ -154,9 +154,13 @@ module.exports = {
...
-          lastVersion: '1.12',
+          lastVersion: '1.13',
           versions: {
             current: {
+              label: '1.14',
+              path: '1.14',
+            },
+            '1.13': {
               label: '1.13',
               path: '1.13',
```
ifbyol, line 165:
```suggestion
              path: '/',
              banner: 'none',
```

You have to change also the `1.12` section to replace the `path` property to be `1.12` instead of `/`. As it is out of the context of the review I cannot do the suggestion but it has to be:

```
'1.12': {
  label: '1.12',
  path: '1.12',
  banner: none
}
```

You can use this PR as reference: https://github.com/okteto/docs/pull/484/files

ifbyol (approved):

PR #504, opened 2023-10-03
Document how to backup and restore an Okteto instance

pchico83:
The platform team has prepared a guide about backing up and restoring an Okteto instance.
This is useful for customers running SH.

### DIFF
--- a/src/content/self-hosted/install/backup-restore.mdx
+++ b/src/content/self-hosted/install/backup-restore.mdx
@@ -1,0 +1,50 @@
+---
+title: Backup and Restore
+description: Backup and Restore your Okteto instance
+id: backup
+sidebar_label: Backup and Restore
+---
+
+Backup and restore planning is essential to ensure the availability and integrity of your Okteto instance in case of unexpected failures or disaster recovery.
+This guide outlines the steps to recover an Okteto instance in such scenarios.
+
+## Critical Resources
+
+To effectively recover an Okteto instance, focus on the following critical resources:
+
+- **Service Accounts**: Preserve service accounts associated with users. These service accounts reside within the `okteto` namespace and are labeled with `dev.okteto.com: "true"`.
+- **Secrets**: Recreate all secrets within the `okteto` namespace.
+- **Custom Resources**: Restore Custom Resources (CRs) corresponding to Okteto's CRDs. They can be identified with the label `app.kubernetes.io/part-of: okteto`.
+
+## Recovery Process
+
+You can follow two different recovery processes:
+
+- **Light recovery processes**: recovers all the users’ metadata, tokens and secrets, but it doesn’t recover users’ namespaces and volumes. All this information is available on the `okteto` namespace.
+- **Full recovery processes**: recovers not only the `okteto` namespace, but also all the users’ namespaces and preview environments, including data in persistent volumes.
+
+### Light Recovery Process
+
+If you do not wish to retain users’ namespaces and their application data, migrating the resources in the `okteto` namespace is sufficient. This is the fastest option, but your users must recreate their development environments.
+
+Follow these steps:
+
+- **Backup the okteto namespace**: backup the [critical resources](self-hosted/administration/configuration.mdx#subdomain) in your `okteto` namespace.
+- **Restore Service Accounts, Secrets, and Custom Resources**: restore service accounts, secrets, and Okteto custom resources in the `okteto` namespace.
+- **Reinstall Okteto**: reinstall the Okteto Helm chart with the same configuration as your previous Okteto instance.
+- **Switch DNS**: the new Ingress-Nginx service will have a new load balancer. Configure your DNS settings to point to the IP of this new load balancer.
+- **Delete the broken cluster**: once the restore work is complete and it has been verified that the new cluster is functioning correctly, proceed to delete the old one.
+
+### Full cluster Recovery Process
+
+If you wish or need to restore all user namespaces, performing the following action before starting the Light Recovery Process guide is necessary.
+
+- **Backup user namespaces**: you can use a tool like [Velero](https://velero.io/) to back up your users’ namespaces. All these namespaces have the label `dev.okteto.com: "true"`.
+- **Backup the okteto namespace**: backup the [critical resources](self-hosted/administration/configuration.mdx#subdomain) in your `okteto` namespace.
+- **Restore user namespaces**: recreate users’s namespaces in the new cluster before reinstalling Okteto.
+- **Restore Service Accounts, Secrets, and Custom Resources**: Restore service accounts, secrets, and Okteto custom resources in the `okteto` namespace.
+- **Reinstall Okteto**: reinstall the Okteto Helm chart with the same configuration as your previous Okteto instance.
+- **Switch DNS**: the new Ingress-Nginx service will have a new load balancer. Configure your DNS settings to point to the IP of this new load balancer.
+- **Delete the broken cluster**: once the restore work is complete and it has been verified that the new cluster is functioning correctly, proceed to delete the old one.
+
+> Keep in mind that private registry configurations are not reestablished until the Helm chart is reinstalled. This may temporarily affect the ability of certain pods to pull images from the Okteto Registry.
--- a/src/content/self-hosted/install/backup-restore.mdx
+++ b/src/content/self-hosted/install/backup-restore.mdx
@@ -5,14 +5,13 @@ id: backup
 sidebar_label: Backup and Restore
 ---
 
-Backup and restore planning is essential to ensure the availability and integrity of your Okteto instance in case of unexpected failures or disaster recovery.
-This guide outlines the steps to recover an Okteto instance in such scenarios.
+This guide outlines the steps to recover an Okteto instance in the event of an unexpected failure and for disaster recovery preparedness.
 
 ## Critical Resources
 
-To effectively recover an Okteto instance, focus on the following critical resources:
+The most important resources to backup so you can effectively recover an Okteto instance are:
 
-- **Service Accounts**: Preserve service accounts associated with users. These service accounts reside within the `okteto` namespace and are labeled with `dev.okteto.com: "true"`.
+- **Service Accounts**: Preserve any service accounts associated with users. These service accounts reside within the `okteto` namespace and are labeled with `dev.okteto.com: "true"`.
 - **Secrets**: Recreate all secrets within the `okteto` namespace.
 - **Custom Resources**: Restore Custom Resources (CRs) corresponding to Okteto's CRDs. They can be identified with the label `app.kubernetes.io/part-of: okteto`.
 
@@ -37,7 +36,9 @@ Follow these steps:
 
 ### Full cluster Recovery Process
 
+:::caution
 If you wish or need to restore all user namespaces, performing the following action before starting the Light Recovery Process guide is necessary.
+:::
 
 - **Backup user namespaces**: you can use a tool like [Velero](https://velero.io/) to back up your users’ namespaces. All these namespaces have the label `dev.okteto.com: "true"`.
 - **Backup the okteto namespace**: backup the [critical resources](self-hosted/administration/configuration.mdx#subdomain) in your `okteto` namespace.
--- a/src/content/self-hosted/install/backup-restore.mdx
+++ b/src/content/self-hosted/install/backup-restore.mdx
@@ -7,45 +7,48 @@ sidebar_label: Backup and Restore
 
 This guide outlines the steps to recover an Okteto instance in the event of an unexpected failure and for disaster recovery preparedness.
 
-## Critical Resources
+## Okteto Metadata
 
 The most important resources to backup so you can effectively recover an Okteto instance are:
 
-- **Service Accounts**: Preserve any service accounts associated with users. These service accounts reside within the `okteto` namespace and are labeled with `dev.okteto.com: "true"`.
-- **Secrets**: Recreate all secrets within the `okteto` namespace.
-- **Custom Resources**: Restore Custom Resources (CRs) corresponding to Okteto's CRDs. They can be identified with the label `app.kubernetes.io/part-of: okteto`.
+- **Service Accounts**: Okteto creates a Kubernetes service account for every user that logins into Okteto. Backup all the service accounts in the `okteto` namespace with the label `dev.okteto.com: "true"`.
+- **Secrets**: backup all the secrets in the `okteto` namespace. Okteto uses secrets to store Okteto Secrets, tokens, certificates, ssh keypairs, ...
+- **Custom Resources**: backup Custom Resources in the `okteto` namespace corresponding to the Okteto's CRDs. Okteto CRDs can be identified with the label `app.kubernetes.io/part-of: okteto`. Okteto uses CRDs to store your Catalog, Container Registry Credentials, ...
+
+We recommend to backup these resources every 24h.
 
 ## Recovery Process
 
 You can follow two different recovery processes:
 
-- **Light recovery processes**: recovers all the users’ metadata, tokens and secrets, but it doesn’t recover users’ namespaces and volumes. All this information is available on the `okteto` namespace.
-- **Full recovery processes**: recovers not only the `okteto` namespace, but also all the users’ namespaces and preview environments, including data in persistent volumes.
+- **Okteto Metadata recovery processes**: recovers all the users’ metadata, tokens, secrets and custom resources. It doesn’t recover users’ namespaces, preview environments and volumes.
+- **Okteto Metadata and Users' Namespaces recovery processes**: recovers not only the Okteto metadata, but also all the users’ namespaces, preview environments, and volumes.
 
-### Light Recovery Process
+Since namespaces are ephemeral in Okteto, we ecommend the first option and let your developers recreate their namespaces as needed.
 
-If you do not wish to retain users’ namespaces and their application data, migrating the resources in the `okteto` namespace is sufficient. This is the fastest option, but your users must recreate their development environments.
+### Okteto Metadata Recovery Process
 
-Follow these steps:
+This is the fastest and easiest recovery process.
+If you do not wish to restore users’ namespaces and their application data, restoring the [Okteto Metadata](self-hosted/install/backup-restore.mdx/#okteto-metadata) is enough:
 
-- **Backup the okteto namespace**: backup the [critical resources](self-hosted/administration/configuration.mdx#subdomain) in your `okteto` namespace.
-- **Restore Service Accounts, Secrets, and Custom Resources**: restore service accounts, secrets, and Okteto custom resources in the `okteto` namespace.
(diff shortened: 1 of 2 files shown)


src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -0,0 +1,50 @@
+---
+title: Backup and Restore
+description: Backup and Restore your Okteto instance
+id: backup
+sidebar_label: Backup and Restore
+---
+
+Backup and restore planning is essential to ensure the availability and integrity of your Okteto instance in case of unexpected failures or disaster recovery.
```
omnomagonz, line 8:
```suggestion
```

Suggesting this out so I can edit only the next line

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -0,0 +1,50 @@
...
+This guide outlines the steps to recover an Okteto instance in such scenarios.
```
omnomagonz, line 9:
```suggestion
This guide outlines the steps to recover an Okteto instance in the event of an unexpected failure and for disaster recovery preparedness.
```

I think we can condense these first two sentences into this one. WDYT @pchico83?

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+## Critical Resources
+
+To effectively recover an Okteto instance, focus on the following critical resources:
```
omnomagonz, line 13:
```suggestion
The most important resources to focus on to effectively recover an Okteto instance are:
```

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+- **Service Accounts**: Preserve service accounts associated with users. These service accounts reside within the `okteto` namespace and are labeled with `dev.okteto.com: "true"`.
```
omnomagonz, line 15:
```suggestion
- **Service Accounts**: Preserve any service accounts associated with users. These service accounts reside within the `okteto` namespace and are labeled with `dev.okteto.com: "true"`.
```

Small suggested change. 

Do we want to explain _why_ these service accounts are so important for recovery? For example: does this save a lot of configuration/setup time? I think including rationale to explain why this specific resource is important may help here

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -0,0 +1,50 @@
...
+- **Secrets**: Recreate all secrets within the `okteto` namespace.
```
omnomagonz, line 16:
```suggestion
- **Secrets**: Recreate all secrets within the `okteto` namespace.
```

The first sentence in this section says these are the most critical resources to "focus on" to effectively recovery an Okteto instance.

The first resource (Service Accounts) seems to be worded in a way that suggests "you should _backup_ any service accounts linked to users".

Now this resource (Secrets) seems to describe what the outcome would be of restoring Secrets, which deviates from the previous resource.

Is the intent here to say: "You should _backup_ Secrets because it will save you from having to recreate all of them when restoring the instance"?

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -0,0 +1,50 @@
...
+- **Custom Resources**: Restore Custom Resources (CRs) corresponding to Okteto's CRDs. They can be identified with the label `app.kubernetes.io/part-of: okteto`.
+
+## Recovery Process
+
+You can follow two different recovery processes:
+
+- **Light recovery processes**: recovers all the users’ metadata, tokens and secrets, but it doesn’t recover users’ namespaces and volumes. All this information is available on the `okteto` namespace.
```
omnomagonz, line 23: @pchico83 I'm wondering if it might be more clear to say something like:

- User only recovery
- User & data recovery

Something that's descriptive of what's being recovered. This seems like the general line of demarcation for what we recover in light vs. full, but WDYT?

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -0,0 +1,50 @@
...
+- **Full recovery processes**: recovers not only the `okteto` namespace, but also all the users’ namespaces and preview environments, including data in persistent volumes.
+
+### Light Recovery Process
+
+If you do not wish to retain users’ namespaces and their application data, migrating the resources in the `okteto` namespace is sufficient. This is the fastest option, but your users must recreate their development environments.
+
+Follow these steps:
+
+- **Backup the okteto namespace**: backup the [critical resources](self-hosted/administration/configuration.mdx#subdomain) in your `okteto` namespace.
```
omnomagonz, line 32:
```suggestion
- **Backup the okteto namespace**: backup the [critical resources](self-hosted/administration/configuration.mdx#subdomain) in your `okteto` namespace.
```

This link on `critical resources` seems odd. Is this intentional and if so, why is the `#subdomain` anchor content critical?

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -0,0 +1,50 @@
...
+- **Restore Service Accounts, Secrets, and Custom Resources**: restore service accounts, secrets, and Okteto custom resources in the `okteto` namespace.
+- **Reinstall Okteto**: reinstall the Okteto Helm chart with the same configuration as your previous Okteto instance.
```
omnomagonz, line 34: I'm confused here. The first step in the light _recovery_ process is to _backup_ the okteto namespace. Shouldn't that happen before a recovery process? I would expect the first step to be _restore_ the okteto namespace that was previously backed up.

Then I'm also not clear on how the restoration works. If there's an unexpected failure of some kind, where are these resources being restored that we can then reinstall Okteto via the helm chart? Are these being restored directly into some data store or something?

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -0,0 +1,50 @@
...
+- **Switch DNS**: the new Ingress-Nginx service will have a new load balancer. Configure your DNS settings to point to the IP of this new load balancer.
+- **Delete the broken cluster**: once the restore work is complete and it has been verified that the new cluster is functioning correctly, proceed to delete the old one.
```
omnomagonz, line 36: I think this answers my question above about _where_ users restore their backups to then reinstall Okteto.

I think that may be a missing detail in the beginning somewhere to say: "Backup these critical resources. When you're ready to restore everything, create a new cluster..."

This also implies (to me) that there's zero prep work to be done on that new cluster. If that's true then we could add a point in the recovery process to create a new cluster.

If there is additional prep work then that warrants its own section to provide that detail, imo

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+### Full cluster Recovery Process
+
+If you wish or need to restore all user namespaces, performing the following action before starting the Light Recovery Process guide is necessary.
```
omnomagonz, line 40:
```suggestion
:::caution
If you wish or need to restore all user namespaces, performing the following action before starting the Light Recovery Process guide is necessary.
:::
```
If this section is required before starting the "light" recovery process then we might want to put this section first. 

I've also turned this into an admonition to be more obvious to callout this requirement.

src/content/self-hosted/install/backup-restore.mdx
omnomagonz: I read that docs page and I think I have a better idea of why you're linking here. Is this ultimately about "backing up" the DNS entry for the desired subdomain used for the namespace? If so, we might just explain that here since it's relatively short/small content rather than link to a section on a bigger page that doesn't "get right to the point" since it's serving a different purpose elsewhere.

WDYT?

src/content/self-hosted/install/backup-restore.mdx
pchico83: @omnomagonz makes sense, I will add an explanation

src/content/self-hosted/install/backup-restore.mdx
pchico83: You are right, I will update based on your suggestion

src/content/self-hosted/install/backup-restore.mdx
pchico83: Right, much more clear. I will update

src/content/self-hosted/install/backup-restore.mdx
pchico83: It's a typo, I wanted to link to the previous section in this doc. Instead of "Critical resources" I am thinking to rename it to "Okteto Metadata", or "Okteto internal data"

src/content/self-hosted/install/backup-restore.mdx
pchico83: It's restored in the previous step:
```
**Restore Service Accounts, Secrets, and Custom Resources**: restore service accounts, secrets, and Okteto custom resources in the `okteto` namespace.
```

src/content/self-hosted/install/backup-restore.mdx
pchico83: Good point, I will add it

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -5,46 +5,50 @@ id: backup
...
+The most important resources to backup so you can effectively recover an Okteto instance are:
 
-- **Service Accounts**: Preserve service accounts associated with users. These service accounts reside within the `okteto` namespace and are labeled with `dev.okteto.com: "true"`.
-- **Secrets**: Recreate all secrets within the `okteto` namespace.
-- **Custom Resources**: Restore Custom Resources (CRs) corresponding to Okteto's CRDs. They can be identified with the label `app.kubernetes.io/part-of: okteto`.
+- **Service Accounts**: Okteto creates a Kubernetes service account for every user that logins into Okteto. Backup all the service accounts in the `okteto` namespace with the label `dev.okteto.com: "true"`.
+- **Secrets**: backup all the secrets in the `okteto` namespace. Okteto uses secrets to store Okteto Secrets, tokens, certificates, ssh keypairs, ...
+- **Custom Resources**: backup Custom Resources in the `okteto` namespace corresponding to the Okteto's CRDs. Okteto CRDs can be identified with the label `app.kubernetes.io/part-of: okteto`. Okteto uses CRDs to store your Catalog, Container Registry Credentials, ...
```
omnomagonz, line 16:
```suggestion
- **Custom Resources**: Any Custom Resources in the `okteto` namespace corresponding to the Okteto CRDs. These CRDs can be identified with the label `app.kubernetes.io/part-of: okteto`. Okteto uses CRDs to store your Catalog, Container Registry Credentials, and more
```

Small changes to match tone of the above two resources. Let me know if this works/you agree with the changes 🙂

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -5,46 +5,50 @@ id: backup
...
+
+We recommend to backup these resources every 24h.
 
 ## Recovery Process
 
 You can follow two different recovery processes:
 
-- **Light recovery processes**: recovers all the users’ metadata, tokens and secrets, but it doesn’t recover users’ namespaces and volumes. All this information is available on the `okteto` namespace.
-- **Full recovery processes**: recovers not only the `okteto` namespace, but also all the users’ namespaces and preview environments, including data in persistent volumes.
+- **Okteto Metadata recovery processes**: recovers all the users’ metadata, tokens, secrets and custom resources. It doesn’t recover users’ namespaces, preview environments and volumes.
```
omnomagonz, line 24:
```suggestion
- **Okteto Metadata recovery processes**: recovers all the users’ metadata, tokens, secrets and custom resources. It doesn’t recover users’ namespaces, preview environments and volumes.
```

Do we have specific examples of what "metadata" is? If it's the `Service Accounts`, `Secrets`, and `Custom Resources` listed above then we can probably leave out the duplicate mentions here. Maybe instead link back up to that Metadata section above?

If "metadata" means something different here then I think we should expand on that and consider renaming "Okteto Metadata" above 😅

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -5,46 +5,50 @@ id: backup
...
+- **Okteto Metadata and Users' Namespaces recovery processes**: recovers not only the Okteto metadata, but also all the users’ namespaces, preview environments, and volumes.
```
omnomagonz, line 25:
```suggestion
- **Okteto Metadata and Users' Namespaces recovery process**: recovers the Okteto metadata, namespaces, preview environments, and volumes for all users. This is a full and complete recovery of the Okteto instance
```

Some suggestions to clarify but making some assertions I'm not sure of. Is this in fact a full recovery?

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -5,46 +5,50 @@ id: backup
...
 
-### Light Recovery Process
+Since namespaces are ephemeral in Okteto, we ecommend the first option and let your developers recreate their namespaces as needed.
```
omnomagonz, line 27:
```suggestion
Namespaces are ephemeral in Okteto so we recommend using the `Okteto Metadata recovery process` and let your developers recreate their namespaces as needed.
```

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -5,46 +5,50 @@ id: backup
...
 
-If you do not wish to retain users’ namespaces and their application data, migrating the resources in the `okteto` namespace is sufficient. This is the fastest option, but your users must recreate their development environments.
+### Okteto Metadata Recovery Process
 
-Follow these steps:
+This is the fastest and easiest recovery process.
```
omnomagonz, line 31:
```suggestion
This is the fastest and easiest recovery process. This recovery process will **not** restore users' namespaces and application data.
```

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -5,46 +5,50 @@ id: backup
...
+If you do not wish to restore users’ namespaces and their application data, restoring the [Okteto Metadata](self-hosted/install/backup-restore.mdx/#okteto-metadata) is enough:
```
omnomagonz, line 32:
```suggestion
```

I think adding part of this to the line above and removing this is sufficient

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -5,46 +5,50 @@ id: backup
...
 
-- **Backup the okteto namespace**: backup the [critical resources](self-hosted/administration/configuration.mdx#subdomain) in your `okteto` namespace.
-- **Restore Service Accounts, Secrets, and Custom Resources**: restore service accounts, secrets, and Okteto custom resources in the `okteto` namespace.
-- **Reinstall Okteto**: reinstall the Okteto Helm chart with the same configuration as your previous Okteto instance.
-- **Switch DNS**: the new Ingress-Nginx service will have a new load balancer. Configure your DNS settings to point to the IP of this new load balancer.
-- **Delete the broken cluster**: once the restore work is complete and it has been verified that the new cluster is functioning correctly, proceed to delete the old one.
+- Backup the [Okteto Metadata](self-hosted/install/backup-restore.mdx/#okteto-metadata) in your `okteto` namespace.
+- Create a new cluster
+- Reinstall Okteto: reinstall the Okteto Helm chart with the same configuration as your previous Okteto instance in the new cluster.
```
omnomagonz, line 36:
```suggestion
- Reinstall the Okteto Helm chart with the same configuration as your previous Okteto instance in the new cluster.
```

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -5,46 +5,50 @@ id: backup
...
+- Restore the [Okteto Metadata](self-hosted/install/backup-restore.mdx/#okteto-metadata) in the `okteto` namespace of the new cluster.
+- Switch DNS: the Ingress-Nginx service in the new cluster will have a new load balancer. Configure your DNS settings to point to the IP of this new load balancer.
```
omnomagonz, line 38:
```suggestion
- The Ingress-Nginx service in the new cluster will have a new load balancer. Configure your DNS settings to point to the IP of this new load balancer.
```

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -5,46 +5,50 @@ id: backup
...
+- Delete the broken cluster: once the restore work is complete and it has been verified that the new cluster is functioning correctly, proceed to delete the old one.
```
omnomagonz, line 39:
```suggestion
- Once the restore work is complete and it has been verified that the new cluster is functioning correctly you can delete the old cluster
```

src/content/self-hosted/install/backup-restore.mdx
pchico83: It refers to the previous data, I will add a link

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -0,0 +1,53 @@
...
+- Backup the [Okteto Metadata](self-hosted/install/backup-restore.mdx/#okteto-metadata) in your `okteto` namespace.
+- Create a new cluster
+- Reinstall the Okteto Helm chart with the same configuration as your previous Okteto instance in the new cluster.
+- Restore the [Okteto Metadata](self-hosted/install/backup-restore.mdx/#okteto-metadata) in the `okteto` namespace of the new cluster.
+- The Ingress-Nginx service in the new cluster will have a new load balancer. Configure your DNS settings to point to the IP of this new load balancer.
+- Once the restore work is complete and it has been verified that the new cluster is functioning correctly you can delete the old cluster
+
+### Users and Namespaces Recovery
+
+This option is slower and requires the usage of external tools to backup all your namespaces:
```
omnomagonz, line 42:
```suggestion
This option requires more time and the use of external tools to backup all your namespaces:
```

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+- Backup users' namespaces: you can use a tool like [Velero](https://velero.io/) to backup your users’ namespaces. All these namespaces have the label `dev.okteto.com: "true"`.
+- Backup the [Okteto Metadata](self-hosted/install/backup-restore.mdx/#okteto-metadata) in your `okteto` namespace.
```
omnomagonz, line 45:
```suggestion
- Backup the [Okteto Metadata](self-hosted/install/backup-restore/#okteto-metadata) in your `okteto` namespace.
```

src/content/self-hosted/install/backup-restore.mdx
```diff
@@ -0,0 +1,53 @@
...
+- Create a new cluster
+- Restore users' namespaces: you can use a tool like [Velero](https://velero.io/) to restore users’s namespaces in the new cluster before reinstalling Okteto.
```
omnomagonz, line 47:
```suggestion
- Restore users' namespaces using a tool like [Velero](https://velero.io/) in the new cluster before reinstalling Okteto.
```

omnomagonz (approved):
A few last non-blocking suggestions

PR #528, opened 2023-11-10
fix: envFiles

andreafalzetti:
`envFiles` is using `h5` instead of `h4` so it's not showing in the right hand side menu


ifbyol (approved):
Could we update it in previous version? At least in `1.14` which is the default one currently?

ifbyol:
> Thanks @ifbyol for making me realise this, the prev versions are incorrect too, so I would like to update them all. Do you see any downside?

No, I don't see any downsides

PR #543, opened 2023-12-04
Reorganized docs sidebar navigation

omnomagonz:
Reorganized the sidebar navigation to create a more intuitive doc structure. Based on internal discussions in [this doc](https://www.notion.so/okteto/Phase-1-Documentation-Structure-and-Style-9e3d239a359b4dd28f87b57675cc6333?pvs=4#71fe1933224749f28e298020f61a2f79)

cc @codyjlandstrom @pchico83

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -8,55 +8,102 @@
 module.exports = {
   "docs": [
     {
-      "Welcome": [
+      "Get started": [
       "overview",
-      "saas-vs-self-hosted",
+      "self-hosted/install/preparation",
       {
         "type": "category",
-        "label": "Getting Started",
+        "label": "Installing Okteto",
         "items": [
-          "getting-started",
-          "using-dev-envs"
+          "self-hosted/install/deployment",
+          {
+            "type": "category",
+            "label": "Cloud provider guides",
+            "items": [
+              "self-hosted/eks",
+              "self-hosted/aks",
+              "self-hosted/civo",
+              "self-hosted/do",
+              "self-hosted/gke"
+            ]
+          },
         ]
       }
     ],
   },
   {
-    "Dev Environments": [
-      "reference/development-environments",
+    "Core concepts": [
       "cloud/okteto-cli",
       {
         "type": "category",
-        "label": "Deploying",
+        "label": "Okteto components",
         "items": [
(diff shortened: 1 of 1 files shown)


sidebars.js
```diff
@@ -8,154 +8,179 @@
...
+          "samples/php",
+          "samples/python",
+          "samples/ruby",
+          "samples/more"
+        ],
+      },
+    ]
+  },
+  {
+    "Okteto manifest": [
```
pchico83, line 93: While this section doesn't grow, should we remove it and move the 2 docs under `core concepts`?

sidebars.js
```diff
@@ -8,154 +8,179 @@
...
+      "Get started": [
       "overview",
-      "saas-vs-self-hosted",
+      "self-hosted/install/preparation",
       {
         "type": "category",
-        "label": "Getting Started",
+        "label": "Installing Okteto",
         "items": [
-          "getting-started",
```
pchico83, line 18: we are missing these two. this is our current getting started guide. it should go right after the SH installation

sidebars.js
omnomagonz: The most important page missing (imo) is the "Configure the manifest", which we don't yet have content for and would come in a later PR.

I think it's worth keeping this at the top-level and prioritizing the content for that missing page since I do think it makes sense why we moved this to the top-level

sidebars.js
```diff
@@ -8,154 +8,180 @@
...
+              "self-hosted/aks",
+              "self-hosted/civo",
+              "self-hosted/do",
+              "self-hosted/gke"
+            ]
+          },
         ]
-      }
+      },
+      "using-dev-envs"
```
pchico83, line 32: I think we are missing this one, which explains how to install and configure the Okteto CLI:
https://www.okteto.com/docs/getting-started/
This way, the journey is:
- Install SH
- Install CLI and configure it
- Deploy your first environment (using our sample app)
-

sidebars.js
pchico83: SGTM

pchico83 (approved):

omnomagonz:
Merging to continue progressing on this project. We will address any feedback in follow-up PRs 😄

PR #544, opened 2023-12-04
Diagnostics docs

maroshii:
Adds documentation for okteto diagnostics. We should GA this until we are OK with the experience.

Pending items:

- [x] Better understanding of the data we are exporting
- [x] Check what sensitive information we may be exposing from the okteto namespace and use [redactors](https://troubleshoot.sh/docs/redact/redactors/) to obfuscate.
- [ ] ~~Provide screenshots of the preflights UI~~

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -144,7 +144,8 @@ module.exports = {
           "self-hosted/manage/okteto-license",
           "self-hosted/manage/backup",
           "self-hosted/manage/custom-resource-definitions",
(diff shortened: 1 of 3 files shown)


codyjlandstrom (commented):
Looks good. Added some quick language tweaks and mentioned that it is read-only in the run description. Do we also want to mention that it doesn't install anything on the cluster or is that a given?

codyjlandstrom (approved):
Looks great

PR #550, opened 2023-12-05
Restructuring 'Okteto Manifest' section

omnomagonz:
This PR includes the following changes for the `Okteto manifest` section:

- Moved `external-resources.mdx` and `secrets.mdx` into a new folder `manifest/`
- Renamed `secrets.mdx` -> `okteto-secrets.mdx`
- Updated/fixed links to these pages
- Added redirects for these pages


sidebars.js
```diff
@@ -93,8 +93,8 @@ module.exports = {
   },
   {
     "Okteto manifest": [
-      "cloud/external-resources",
-      "cloud/secrets"
+      "manifest/external-resources",
+      "manifest/okteto-secrets"
```
jmacelroy, line 97: Looking at the preview it seems odd to put this secrets documentation under a section titled `Okteto manifest`. The document is all about adding secrets via the UI and I don't see anything related to manifests. The external resources one is better and I think it's a good fit for an `Okteto manifest` section but there are UI components to the doc as well.

Should we consider renaming the section from "Okteto manifest" if it's going to contain docs unrelated to the manifest OR is the secrets doc an outlier from your plan and it may need a new home?

pchico83: I guess the plan is to update the content later and explain how to refer to your secrets in the okteto manifest.
Btw, not related to this PR, but we call it secrets in some places (admin and personal secrets), and variables in other places (variables in the catalog item or when deploying a repo from the UI). And they all are the same and behave the same. Should we switch to `variables` everywhere?

pchico83: There is an open issue to allow the definition of `variables` i the okteto manifest itself, which would make this subsection more aligned with being part of the okteto manifest

jmacelroy (approved):

PR #554, opened 2023-12-07
Restructuring the 'References' section

omnomagonz:
This PR includes the following changes for the `References` section:

- Renamed the following ids and file names:
  - `cli` -> `okteto-cli`
  - `manifest` -> `okteto-manifest`
  - `compose` -> `docker-compose`
- Added redirects
- Updated links for these new names


omnomagonz:
cc @huguestennier for redirect review (should be quick/simple 🙏 )

cc @jmacelroy for overall review

sidebars.js
```diff
@@ -166,9 +166,9 @@ module.exports = {
   },
   {
     "References": [
-      "reference/cli",
-      "reference/manifest",
-      "reference/compose",
+      "reference/okteto-cli",
+      "reference/okteto-manifest",
+      "reference/docker-compose",
       "reference/helm-chart-settings",
```
pchico83, line 172: I would go with `helm-chart-values` here, `values.yml` is the standard name of this file

PR #556, opened 2023-12-11
Remove references to cluster.endpoint

codyjlandstrom:
This pr removes references to cluster.endpoint in our q
uickstart guide, as well as our cloud provider guides as this is deprecated in 1.16

### DIFF
--- a/src/content/get-started/cloud-provider-guides/aks.mdx
+++ b/src/content/get-started/cloud-provider-guides/aks.mdx
@@ -35,13 +35,6 @@ We recommend the following specs:
 - A pool with at least 3 `Standard D4` nodes
 - 250 GB per Standard SSD Managed Disk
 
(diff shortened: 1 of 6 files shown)


pchico83 (approved):
@codyjlandstrom I think we need to update these files too:
https://github.com/okteto/docs/blob/main/src/content/self-hosted/eks/config.yaml
there are a few of them, for eks, also, gke, civo, ...

codyjlandstrom:
> @codyjlandstrom I think we need to update these files too: https://github.com/okteto/docs/blob/main/src/content/self-hosted/eks/config.yaml there are a few of them, for eks, aks, gke, civo, ...

Thanks for flagging, will update those too.
I also noticed that it is referenced in the example yaml file for each cloud provider (https://www.okteto.com/docs/self-hosted/gke/config.yaml) Do you know where to update those @pchico83?

PR #559, opened 2023-12-12
Move Oketeto Insights to Reference Section

codyjlandstrom:
Move Okteto Insights out from the self-hosted section since this will be available to all users


pchico83:
@codyjlandstrom the whole "Admin features" section is available for SaaS customers. I don't think this is a "reference" doc, since it's explaining a feature. I think we need to find a better place for "Admin features". Maybe a first-level section as it is today?

codyjlandstrom:
> I think we need to find a better place for "Admin features". Maybe a first-level section as it is today?

Yes, now that I'm looking at that broader section, I feel that all of "Admin Features" should be moved out from under "Self-Hosted." What do you think @omnomagonz?

ifbyol:
Agree with creating a first level administration or similar for those features. I think that Okteto Self-Hosted section should host only the things that are specific for SH: installation, upgrade, manage instance (backups, restore, troubleshooting, etc..), license.

omnomagonz:
Yea, +1 to bringing the `admin` section up to top-level. Since it applies to both SaaS and Self-Hosted I think that makes the most sense here

codyjlandstrom:
Agree @omnomagonz, @ifbyol, @pchico83. Bumped out [in the Notion doc](https://www.notion.so/okteto/Phase-1-Documentation-Structure-and-Style-9e3d239a359b4dd28f87b57675cc6333?pvs=4#c4fc1a4edecc493d8f89f7730f692460). If all looks good there, I'll make those changes here

omnomagonz:
@codyjlandstrom I think the `Resource cleanup` (cleanup.mdx) may be missing and we should add that but otherwise LGTM 👍

netlify.toml
```diff
@@ -233,6 +233,26 @@
   to = "/docs/cloud/okteto-cli"
   status = 301
 
+[[redirects]]
```
pchico83, line 236: Instead of adding new redirects, should we modify the existing ones?
https://github.com/okteto/docs/blob/main/netlify.toml#L56

codyjlandstrom: Yes for sure, already working on another commit for that. Wanted to get a preview environment to test since I'm having trouble getting it running locally.

codyjlandstrom: @pchico83 Ready for review

pchico83 (approved):

PR #562, opened 2023-12-14
Add storage driver requirements to EKS install guide

codyjlandstrom:
Add a step to the EKS SH install guide to add the Amazon EBS CLI storage driver add-on

### DIFF
--- a/src/content/get-started/cloud-provider-guides/eks.mdx
+++ b/src/content/get-started/cloud-provider-guides/eks.mdx
@@ -88,6 +88,16 @@ ingress-nginx:
     service:
       annotations:
         service.beta.kubernetes.io/aws-load-balancer-type: nlb
+
+buildkit:
+  persistence:
+    enabled: true
+
+registry:
+  storage:
+    filesystem:
+      persistence:
+        enabled: true
 ```
 
 The `service.beta.kubernetes.io/aws-load-balancer-type: nlb` annotation is used to tell AWS to create a [Network Load Balancer (NLB)](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html) instead of the default [Classic Load Balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html) to expose the ingress controller. We recommend to use a NLB as it provides benefits like: websocket support, static IPs, and better performance. You can see a full comparison between the load balancers [here](https://aws.amazon.com/elasticloadbalancing/features/).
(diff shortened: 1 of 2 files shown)


rberrelleza (approved):

src/content/get-started/cloud-provider-guides/eks.mdx
```diff
@@ -54,6 +54,14 @@ helm repo add okteto https://charts.okteto.com
 helm repo update
 ```
 
+### Configure the Amazon EBS CSI storage driver
```
pchico83, line 57: If we assume the storage driver is available, the user can use this configuration from the beginning:
```
buildkit:
  persistence:
    enabled: true

registry:
  storage:
    filesystem:
      persistence:
        enabled: true
```
Moreover, should we default to persistence for buildkit and the registry in our helm values file?

codyjlandstrom: < Moreover, should we default to persistence for buildkit and the registry in our helm values file?
That makes sense to me. Should we call that out specifically here or just add it to the example file?

pchico83: @codyjlandstrom I think there are two action items here:
- Change the helm values file in this guide as part of this PR. [here](https://github.com/okteto/docs/blob/main/src/content/self-hosted/eks/config.yaml) and [here](https://github.com/okteto/docs/blob/main/src/content/get-started/cloud-provider-guides/eks.mdx#L74)
- Create a jira ticket to update our helm values default file in the chart, which would apply to all the installations. That will enable these values by default across all installations:
```
buildkit:
  persistence:
    enabled: true

registry:
  storage:
    filesystem:
      persistence:
        enabled: true
```
and we could remove them from all the guides. If we do that, we should return a clear error when running `helm install` and a storage driver is unavailable. Maybe with a helm pre-install hook. And remove the step to configure buildkit persistency from the Okteto UI installation dashboard

codyjlandstrom: @pchico83 Updated the PR and created a Jira ticket here: https://okteto.atlassian.net/browse/RAIN-131

pchico83 (approved):

src/content/get-started/cloud-provider-guides/eks.mdx
pchico83: Thanks @codyjlandstrom!

PR #566, opened 2023-12-15
Adding a note about SaaS

omnomagonz:
Adding a note to help SaaS prospects/customers skip unnecessary content.


codyjlandstrom (approved):
Yeah that's a nice add!

PR #567, opened 2023-12-15
Add instructions for retrieving insights bearer token from UI

codyjlandstrom:
Add instructions to the Okteto Insights guide on where to find your Insights bearer token from within the Admin UI


src/content/admin/okteto-insights.mdx
```diff
@@ -15,6 +15,8 @@ This feature is `disabled` by default. To `enable` it you need to set the helm s
 
 Once you enable the feature, Okteto will create a new endpoint accessible through `https://okteto.SUBDOMAIN/metrics` where data will be available in Prometheus format. This new endpoint is protected with a bearer token which is auto-generated by Okteto. Alternatively, you can specify your own token using the secret defined in this [helm setting](reference/helm-chart-values.mdx/#insights).
 
+To retrieve your Insights bearer token from the Okteto Admin UI, navigate to `Admin` -> `General` and scroll to the `Settigns` section. You can copy your bearer token by clicking on the `Show Insights Token` button in the Insights panel.
```
pchico83, line 18: Should we add a screenshot here?

pchico83 (approved):

src/content/admin/okteto-insights.mdx
codyjlandstrom: I was thinking about that, but it felt a bit excessive to me. I'd like to add screenshots of an example Grafana dashboard soon and think that would be a better use of the page space. What do you think?

PR #569, opened 2023-12-18
Release notes for 1.16

maroshii:
What it says on the tin.

Still to do:

- Versioned docs
- Run Algolia crawler


src/content/release-notes.mdx
```diff
@@ -5,6 +5,28 @@ sidebar_label: Release notes
...
+- Registry Credentials cronjob only update dockerconfig secret if the credentials change.  <!-- 7474 -->
+- Okteto Daemon now update kubelet docker config.json file in nodes reactively when the static docker config secret changes. Bumps okteto/daemon to 0.5.1
+- Fix network policies configuration to limit buildkit access to the nginx controller  <!-- 7505 -->
+- Add Okteto Insights Token to the Admin UI  <!-- 7473 -->
+- Remove customer data from mixpanel events  <!-- 7390 -->
+
+### Features
+
+- Add `okteto_usage_build_duration_seconds` build metrics to okteto insights. More info [here](https://www.okteto.com/docs/1.16/admin/okteto-insights)  <!-- 7445 -->
+- Okteto now runs in ARM architecture
```
pchico83, line 24: @maroshii can we mark this as beta?

pchico83 (approved):

src/content/release-notes.mdx
```diff
@@ -17,6 +17,7 @@ id: release-notes
 - Fix network policies configuration to limit buildkit access to the nginx controller  <!-- 7505 -->
 - Add Okteto Insights Token to the Admin UI  <!-- 7473 -->
 - Remove customer data from mixpanel events  <!-- 7390 -->
+- Update structure of the Okteto Documentaion to make content easier to navigate. All changes [here](https://github.com/okteto/docs/pulls?q=is%3Apr+label%3Adoc-restructure+)
```
codyjlandstrom, line 20: @omnomagonz Added a small note about structure changes to the release notes.

maroshii: I can see why we would want to add this here, although I'm not entirely sure it belongs to the release notes (as opposed to a blog post for eg)

Note: There's a typo in `Documentaion`

maroshii: It's just that this is not strictly part of the release. I mean, the docs are for the Chart, the CLI, Cloud, manifest, samples, etc so the scope of the change is larger in principle than the 1.16 release.
Not totally against it but I would like to get more input here since we haven't done this before afaik.

src/content/reference/helm-chart-values.mdx
```diff
@@ -510,13 +510,15 @@ The garbage collector service. It automatically scales idle applications to zero
 - `scaleToZeroPeriod`: The duration, in hours, that an application or resource must be idle before the garbage collector scales it to zero. Set to zero to disable.
 - `deleteNamespacePeriod`: The duration, in days, that a namespace must be idle before the garbage collector deletes it. Set to zero to disable.
 - `slackWebhook`: If set, the garbage collector will send a notification when it scales a resource to zero or when it deletes a namespace.
+- `upSessionByLastSyncedFile`: If true, the garbage collector will mark okteto up sessions as ready for GC if no file has been synchronized in 1 hour. Default behaviour is to check for an active connection (regardless of inactivity)
```
pchico83, line 513:
```suggestion
- `upSessionByLastSyncedFile`: If true, the garbage collector will mark okteto up sessions as inactive for GC if no file has been synchronized. The default behaviour is to check for an active okteto up session (regardless of files being synched or not)
```
@maroshii about the 1h note, is it 1h? or the time configured in the GC? I guess it's the latter, if the GC period id 4h you need to be 4h without synching files to sleep the namespace, right?

src/content/release-notes.mdx
```diff
@@ -5,6 +5,29 @@ sidebar_label: Release notes
 id: release-notes
 ---
 
+## 1.16.0
+
+20 December 2023
+
+### Improvements
```
pchico83, line 12: Can we standarize in the order of the sections? What we have used more times is:
- Kubernetes versions supported
- Breaking changes
- Deprecations
- Features
- Improvements
- Bugfixes
Also, we are supporting k8s 1.28 now, right?

src/content/release-notes.mdx
```diff
@@ -5,6 +5,29 @@ sidebar_label: Release notes
...
+
+- Okteto garbage collector cronjob now resolves buildkit and ingress hosts to their internal IPs  <!-- 7432 -->
+- Registry Credentials cronjob only update dockerconfig secret if the credentials change.  <!-- 7474 -->
+- Okteto Daemon now update kubelet docker config.json file in nodes reactively when the static docker config secret changes. Bumps okteto/daemon to 0.5.1
+- Fix network policies configuration to limit buildkit access to the nginx controller  <!-- 7505 -->
+- Add Okteto Insights Token to the Admin UI  <!-- 7473 -->
+- Remove customer data from mixpanel events  <!-- 7390 -->
```
pchico83, line 19: Can we skip this one? We aren't sending mixpannel events in SH, I don't want to give the impression that we had a security leak here

src/content/release-notes.mdx
pchico83: @maroshii although it's not ideal, I think it's a good place to mention it because we bump the doc versions with each chart release.
@codyjlandstrom but I think the comment is insufficient. At a minimum, I would mention the new location for "release notes" and the Chart values reference, which are commonly accessed by our customers. And a blog post would be ideal, but not sure if we have capacity to do it

src/content/release-notes.mdx
ifbyol: Yes, we start supporting `1.28` as part of this release, and also, this is the first release where we start supporting 4 k8s versions,, right?

src/content/release-notes.mdx
```diff
@@ -5,6 +5,29 @@ sidebar_label: Release notes
...
+
+### Features
+
+- Add `okteto_usage_build_duration_seconds` build metrics to okteto insights. More info [here](https://www.okteto.com/docs/1.16/admin/okteto-insights)  <!-- 7445 -->
+- When telemetry is enabled, Okteto now exports cluster configuration to Mixpanel on every helm upgrade.
+- A diagnostics tool is now provided to help troubleshoot installations. More info [here](https://www.okteto.com/docs/1.16/self-hosted/manage/diagnostics/)
+- The Okteto CLI used for Pipeline installation has been upgraded to version [2.23.1](https://github.com/okteto/okteto/releases/tag/2.23.1)
+- Move garbage collection of `okteto up` based on last synced file behind a feature flag. More info [here](https://www.okteto.com/docs/reference/helm-chart-values/#gc)   <!-- 7424 -->
```
ifbyol, line 28: The link seems broken, at least I get a not found page. Given that is a link for the documentation, shouldn't it be a relative link to the helm chart configuration page?

src/content/release-notes.mdx
```diff
@@ -5,6 +5,29 @@ sidebar_label: Release notes
...
+
+### Improvements
+
+- Okteto garbage collector cronjob now resolves buildkit and ingress hosts to their internal IPs  <!-- 7432 -->
+- Registry Credentials cronjob only update dockerconfig secret if the credentials change.  <!-- 7474 -->
+- Okteto Daemon now update kubelet docker config.json file in nodes reactively when the static docker config secret changes. Bumps okteto/daemon to 0.5.1
+- Fix network policies configuration to limit buildkit access to the nginx controller  <!-- 7505 -->
+- Add Okteto Insights Token to the Admin UI  <!-- 7473 -->
+- Remove customer data from mixpanel events  <!-- 7390 -->
+- Updated documentation structure to make content easier to navigate and find. Release notes can now be found at [https://www.okteto.com/docs/release-notes/](https://www.okteto.com/docs/release-notes/), and chart values reference can be found at [https://www.okteto.com/docs/reference/helm-chart-values/](https://www.okteto.com/docs/reference/helm-chart-values/).  All documentation changes are listed [here](https://github.com/okteto/docs/pulls?q=is%3Apr+label%3Adoc-restructure+).
```
ifbyol, line 20: Should we use internal links for pages within the docs?

ifbyol (commented):
Shouldn't we be using internal (relative) links for the pages in our docs as we have been doing lately?

src/content/reference/helm-chart-values.mdx
maroshii: It's hardcoded to 1h. I don't see why we would need to align this with the gc schedule. It can be maintained independently. We could add an extra config to configure it but it was not in the requirement. @pchico83 let me know if this is a blocker

src/content/release-notes.mdx
maroshii: This is because of the change in the sidebar. Once we change the default this link will work. Some of the ones below will point to the old paths but we have redirection in place so it should be ok.

maroshii: Btw, I'll change all to relative links

maroshii:
Docusaurus fails to build because of broken links. This is because we are referencing the root path which still points to 1.15 and in 1.16 those links don't exist. Adding versioned docs here to this PR to try and fix the issue

src/content/release-notes.mdx
maroshii: I think it's because of the new sidebar and reorg. See: https://github.com/okteto/docs/pull/569#issuecomment-1862885741

ifbyol: That might be. If not, we can ask Hugues about it, but it makes sense what you mention in the comment. I remember I had some issue in the past for not including the extension in the links, that's why I mentioned, but I don't remember the details

src/content/release-notes.mdx
omnomagonz: @pchico83 I like those suggestions/callouts 👍 

Cody and I can draft a blog post but I suspect it'll be a January publish date given upcoming holidays/PTO

pchico83: @omnomagonz I think the video/text message is fine. My 2$ is that a blogpost isn't worth the investment

omnomagonz: That sounds _extra_ good to me then 😄

PR #572, opened 2023-12-28
Rename launch → deploy

viddo:
Updates wording to _"**Deploy** Dev Environment"_ from prev _"**Launch** …"_.

Note that there are still [more instances that use the word "launch"](https://github.com/search?q=repo%3Aokteto%2Fdocs%20launch&type=code) but more so in the general sense than specifically related to the rename related to the _Deploy dialog_ specifically, so left them as-is. Please advice if you think we should also update such sentences.

For new screenshots in affected docs applied a `-v1.17+` suffix in the filename, to indicate that they are intended from v1.17 and forward and not break older versioned and archived docs.

To see difference with prev docs, compare with [existing images (w/o suffix)](https://github.com/okteto/docs/tree/nicklas/LAKE-145/rename-launch-deploy/static/img) or review preview env with live docs, e.g. [/deploy/launch-from-catalog/](https://www.okteto.com/docs/deploy/launch-from-catalog/) and sibling docs.

Updated redirects and current v1.16 docs files/and URLs to also match and work for next/1.17 version docs. It's necessary because Netlify and the redirects doesn't have any notion of our versioning mechanism, and otherwise would break when we switch from v1.16 → v1.17 for current/default docs version.

[LAKE-145]: https://okteto.atlassian.net/browse/LAKE-145


viddo:
Note that need to switch to v1.17 to see new screenshots in use, e.g. https://deploy-preview-572--okteto-docs.netlify.app/docs/1.17/deploy/deploy-from-catalog/

viddo:
BTW, heads up; I'll be on vacation all of next week/1st week of January - feel free to modify and merge PR before I'm back (Mon Jan 8th)

src/content/deploy/deploy-from-catalog.mdx
```diff
@@ -59,13 +59,13 @@ A dialog will open where you can modify the branch, variables, and path to the O
...
   <Image
-    src={require("@site/static/img/redeploy-catalog.png").default}
+    src={require("@site/static/img/redeploy-catalog-v1.17+.png").default}
     alt="Redeploy from the catalog"
     width="600"
   />
 </p>
 
-Click the `Relaunch` button to rebuild and redeploy your application.
+Click the `Redploy` button to rebuild and redeploy your application.
```
omnomagonz, line 68:
```suggestion
Click the `Redeploy` button to rebuild and redeploy your application.
```

omnomagonz (approved):
Looks good to me. Made one minor change to fix a typo. Otherwise looks ready to ship 👍

PR #574, opened 2024-01-03
docs: OKTETO_COMPOSE_VOLUME_AFFINITY_ENABLED

andreafalzetti:


pchico83 (approved):

src/content/reference/docker-compose.mdx
```diff
@@ -387,6 +387,8 @@ volumes:
 Host volumes are added to your containers at deployment time.
 In addition, `okteto up` uses host volumes to infer the [sync paths](reference/okteto-manifest.mdx#sync-string-required) between your local filesystem and your remote development container.
 
+By default, services mounting the same volume will be placed on the same node using Kubernetes's pod affinity. If you wish to disable this behaviour, set `OKTETO_COMPOSE_VOLUME_AFFINITY_ENABLED` to `false`.
```
rberrelleza, line 390: @andreafalzetti could we specify where this should be added? Is this on the command, on the compose? 

Could this be added as an Okteto secret or catalog configuration?

andreafalzetti, line 390:
```suggestion
By default, services mounting the same volume will be placed on the same node using Kubernetes's pod affinity. If you wish to disable this behavior, set the environment variable `OKTETO_COMPOSE_VOLUME_AFFINITY_ENABLED` to `false` when deploying.
```

andreafalzetti: @rberrelleza good point, I am biased because I have implemented it so I know where to specify it, but it wasn't actually clarified. It should be an env var of the `deploy` so either from local or via the UI. Do think think this suggestion clarify it enough?

omnomagonz: @andreafalzetti I think this copy works if the recommendation is to set this variable in the manifest itself (which I assume is what you mean by 'local'). If there's no real recommendation then we could consider adding something like:

```suggestion
By default, services mounting the same volume will be placed on the same node using Kubernetes's pod affinity. If you wish to disable this behavior, set the environment variable `OKTETO_COMPOSE_VOLUME_AFFINITY_ENABLED` to `false` when deploying. You can also optionally configure this variable from the Okteto UI to make it available to all development environments in your Okteto instance.
```

andreafalzetti: Thanks, @omnomagonz!

andreafalzetti: Resolved with @omnomagonz in 48629cb48d9b100f27ed008cbd5709dffb51bd3e

PR #586, opened 2024-01-12
add docs for change namespace owner

codyjlandstrom:
Add the documentation for the change namespace owner feature. Includes a small cleanup of the namespace component page


src/content/core/components/namespaces.mdx
```diff
@@ -23,6 +23,21 @@ Switch your kubectl configuration to point to your namespaces with the `okteto k
 
 Once you're done, you can also delete your namespace via the UI, or with the `okteto namespace delete` command.
 
+You can sleep your namespace from the command line with `okteto namespace sleep [name]`.
+
+When you want to wake your namespace, you can use the command `okteto namespace wake [name]`.
+
+You can sleep your namespace from the Okteto dashboard by clicking on the Sleep dialog.
```
pchico83, line 30:
```suggestion
You can sleep your namespace from the Okteto UI by clicking on the Sleep dialog.
```

src/content/core/components/namespaces.mdx
```diff
@@ -31,18 +46,27 @@ To share a namespace go to the Okteto dashboard, select the namespace you want t
 
 <p align="center"><Image src={require('@site/static/img/share.png').default} alt="share a namespace" width="850" /></p>
 
-In the `Share` dialog, type the email address of the team members you want to share this namespace with. Once you press save, Okteto dashboardll update the necessary permissions and notify your teammates via email.
+In the `Share` dialog, type the email address of the team members you want to share this namespace with. Once you press save, Okteto dashboard will update the necessary permissions and notify your teammates via email.
```
pchico83, line 49:
```suggestion
In the `Share` dialog, type the email address of the team members you want to share this namespace with. Once you press save, Okteto will update the necessary permissions and notify your teammates via email.
```

src/content/core/components/namespaces.mdx
```diff
@@ -31,18 +46,27 @@ To share a namespace go to the Okteto dashboard, select the namespace you want t
...
 
-You can sleep your namespace from the command line with `okteto namespace sleep [name]`.
 
-When you want to wake your namespace, you can use the command `okteto namespace wake [name]`.
+## Transfer your namespace
+
+Okteto administrators have the ability to transfer non-default namespaces between users. This process can be achieved through the Okteto dashboard.
```
pchico83, line 54:
```suggestion
Okteto administrators have the ability to transfer non-default namespaces between users. This process can be achieved through the Okteto UI.
```

src/content/core/components/namespaces.mdx
```diff
@@ -31,18 +46,27 @@ To share a namespace go to the Okteto dashboard, select the namespace you want t
...
+
+To transfer a namespace to a new owner, go to the Okteto dashboard -> Admin -> Namespaces then locate the namespace that you would like to transfer. Click on the three dots `...` to display the namespace menu. From the namespace menu select Change Owner.
```
pchico83, line 56:
```suggestion
To transfer a namespace to a new owner, go to the Okteto UI -> Admin -> Namespaces then locate the namespace that you would like to transfer. Click on the three dots `...` to display the namespace menu. From the namespace menu select `Change Owner`.
```

src/content/core/components/namespaces.mdx
```diff
@@ -31,18 +46,27 @@ To share a namespace go to the Okteto dashboard, select the namespace you want t
...
+If you wish, you can keep the current owner as a member of the namespace by selecting the option in the Change Owner modal. If this is not selected the current owner will immediately lose access.
+
+<p align="center"><Image src={require('@site/static/img/namespace-transfer-modal.png').default} alt="share a namespace" width="700" /></p>
 
-In order to wake the namespace, you can click on Wake dialog.
 
 :::tip
 
-SaaS single tenant and Self-Hosted customers can specify a global, default threshold for automatically sleeping a namespace in the [garbage collection](reference/helm-chart-values.mdx/#gc) configuration.
+User specific secrets will not transfer along with a namespace. If the namespace being transferred has user specific secrets, please note that future build or destroy actions may fail.
```
pchico83, line 68:
```suggestion
User specific variables will not transfer along with a namespace. If the namespace being transferred has user specific variables, please note that future build or destroy actions may fail.
```

src/content/core/components/namespaces.mdx
```diff
@@ -31,18 +46,27 @@ To share a namespace go to the Okteto dashboard, select the namespace you want t
...
 
-:::
+New user specific secrets can be added to a transferred namespace if desired to avoid future failures.
```
pchico83, line 70:
```suggestion
New user specific variables can be added to a transferred namespace if desired to avoid future failures.
```

pchico83 (approved):
I think we are using Okteto UI instead of Okteto dashboard in other docs (cc: @omnomagonz)

src/content/core/components/namespaces.mdx
```diff
@@ -31,18 +46,27 @@ To share a namespace go to the Okteto dashboard, select the namespace you want t
...
+## Transfer your namespace
+
+Okteto administrators have the ability to transfer non-default namespaces between users. This process can be achieved through the Okteto dashboard.
+
+To transfer a namespace to a new owner, go to the Okteto dashboard -> Admin -> Namespaces then locate the namespace that you would like to transfer. Click on the three dots `...` to display the namespace menu. From the namespace menu select Change Owner.
 
-You can sleep your namespace from Okteto dashboardself by clicking on Sleep dialog.
+<p align="center"><Image src={require('@site/static/img/namespace-change-owner.png').default} alt="share a namespace" width="650" /></p>
+
+When the Change Owner modal appears, type the email address of the team member who you'd like the transfer the namespace to. Once you press confirm, ownership of that namespace will be immediately transferred.
```
bgonp, line 60: I would name the modal `Transfer Ownership` since it's the actual title of the modal, rather than the option name in the dropdown menu.
```suggestion
When the Transfer Ownership modal appears, type the email address of the team member who you'd like the transfer the namespace to. Once you press confirm, ownership of that namespace will be immediately transferred.
```

src/content/core/components/namespaces.mdx
bgonp, line 70: I think a link here would be helpful, to take the user _somewhere_ in our docs that lets the user know how to add variables.

src/content/core/components/namespaces.mdx
ifbyol: > If the namespace being transferred has user specific variables

The namespace doesn't have variables. The users are the one having variables and then, they can be used in deployed dev environments within that namespace.

So, the right sentence would be something like: `If previous owner had some user variables used to deploy some active dev environment, please note that future build, redeploy or destroy actions may fail`

src/content/core/components/namespaces.mdx
ifbyol: Same comment about "variables added to a transferred namespace". Namespaces don't have variables.

To avoid future issues, the best way would be to make use of cluster variables or specify the value at deployment time through the UI or with `--var` parameter in the CLI. 

I think we need to think about the future of user specific variables. They are problematic for sharing namespace, transfer ownership or admin performing actions in any namespace scenarios

src/content/core/components/namespaces.mdx
omnomagonz: I'm generally supportive of this suggested change. Why did we opt for two different names though between the menu and the modal? Wouldn't it make sense to keep it the same?

Not a blocking question/suggestion, just curious and to know if we could consider an update in future releases 😄

src/content/core/components/namespaces.mdx
omnomagonz: Agreed. We can use `manifest/okteto-variables.mdx`

omnomagonz: I applied this change though I'm not sure if I should have 😅 

@ifbyol is your comment in opposition to this change given the dynamic you're referring to?

src/content/core/components/namespaces.mdx
codyjlandstrom: It looks like there was a discrepancy in the design file. The original said "Change Owner" and an update said "Transfer Ownership." My bad I should have caught that.
@bgonp is to to late to change the modal title to "Change Owner" to match the dropdown menu?

bgonp: The 1.17 cut has been done but we can create a quick backport for this thing.
Just a note: AFAIK internally both backend and fronted use the "transfer ownership" term. Do we want to take the "Change Owner" as the _official_ name everywhere or "Transfer Ownership"?

codyjlandstrom: I believe "Transfer Ownership" is concise and more accurately describes the feature.
Could we standardize on that? I'll go ahead and update all documentation.

src/content/core/components/namespaces.mdx
ifbyol: Rename from secrets to variable is fine. My comment is more about the fact of adding variables to a namespace. That cannot be done. The namespace cannot have variables

codyjlandstrom: Thanks for the clarification. How would you rephrase the final section of the Tip? Should that be removed all together?

ifbyol: instead of 

```
User specific variables will not transfer along with a namespace. If the namespace being transferred has user specific variables, please note that future build or destroy actions may fail.

New user specific variables can be added to a transferred namespace if desired to avoid future failures. Details on [adding variables can be found here](manifest/okteto-variables.mdx).
```

We could write something like:

```
User specific variables will not transfer along with a namespace. If some deployed development environment within the transferred namespace was deployed using user specific variables, please note that future build or destroy actions may fail.

New user specific variables can be added to the new owner. Details on [adding variables can be found here](manifest/okteto-variables.mdx). Also, consider to move user specific variables to cluster variables (link to cluster variables)
```

Also, I think it might be relevant to highlight the fact that has to be the new owner the one that has to add the new variables to themselves. The admin transferring the namespace cannot add user variables to other user different than themselves

src/content/core/components/namespaces.mdx
```diff
@@ -31,18 +46,27 @@ To share a namespace go to the Okteto dashboard, select the namespace you want t
...
 
-In the `Share` dialog, type the email address of the team members you want to share this namespace with. Once you press save, Okteto dashboardll update the necessary permissions and notify your teammates via email.
+In the `Share` dialog, type the email address of the team members you want to share this namespace with. Once you press save, Okteto will update the necessary permissions and notify your teammates via email.
 
-You can sleep your namespace from the command line with `okteto namespace sleep [name]`.
 
-When you want to wake your namespace, you can use the command `okteto namespace wake [name]`.
+## Transfer namespace ownership
+
+Okteto administrators have the ability to transfer non-default namespaces between users. This process can be achieved through the Okteto UI.
```
maroshii, line 54: Is there somewhere we can link `non-default` namespaces to? Not sure this is a widely understood concept for users/admins

src/content/core/components/namespaces.mdx
```diff
@@ -31,18 +46,27 @@ To share a namespace go to the Okteto dashboard, select the namespace you want t
...
+If you wish, you can keep the current owner as a member of the namespace by selecting the option in the Transfer Ownership modal. If this is not selected the current owner will immediately lose access.
+
+<p align="center"><Image src={require('@site/static/img/namespace-transfer-modal.png').default} alt="select a new namespace owner" width="700" /></p>
 
-In order to wake the namespace, you can click on Wake dialog.
 
 :::tip
 
-SaaS single tenant and Self-Hosted customers can specify a global, default threshold for automatically sleeping a namespace in the [garbage collection](reference/helm-chart-values.mdx/#gc) configuration.
+User specific variables will not transfer along with a namespace. If the namespace being transferred has user specific variables, please note that future build or destroy actions may fail.
```
maroshii, line 68: I would change `:::tip` for `:::note` here

src/content/core/components/namespaces.mdx
```diff
@@ -23,6 +23,21 @@ Switch your kubectl configuration to point to your namespaces with the `okteto k
 
 Once you're done, you can also delete your namespace via the UI, or with the `okteto namespace delete` command.
 
+You can sleep your namespace from the command line with `okteto namespace sleep [name]`.
+
+When you want to wake your namespace, you can use the command `okteto namespace wake [name]`.
+
+You can sleep your namespace from the Okteto UI by clicking on the Sleep dialog.
+
+In order to wake the namespace, you can click on the Wake dialog.
```
rberrelleza, line 32: Would it be worth having a section on awake and sleep and explaining what each is, and including the information here (plus the other ways to awake & sleep). I know this is out of the scope of the change, so feel free to ignore or postpone.

src/content/core/components/namespaces.mdx
```diff
@@ -23,6 +23,21 @@ Switch your kubectl configuration to point to your namespaces with the `okteto k
...
+
+:::tip
+
+SaaS single tenant and Self-Hosted customers can specify a global, default threshold for automatically sleeping a namespace in the [garbage collection](reference/helm-chart-values.mdx/#gc) configuration.
```
rberrelleza, line 36: Now that cloud is out, we only have single-tenant instances, so we can simplify the instructions: 

```suggestion
Administrators can specify a global, default threshold for automatically sleeping a namespace in the [garbage collection](reference/helm-chart-values.mdx/#gc) configuration.
```

src/content/core/components/namespaces.mdx
codyjlandstrom: Good call, Thanks

src/content/core/components/namespaces.mdx
codyjlandstrom: Yes, agreed. I feel like information on default namespaces should be covered in this page. The namespace management section at the top really needs an overhaul. I'm going to make a comment to handle that work separate from this PR.

src/content/core/components/namespaces.mdx
codyjlandstrom: Yes, agreed. I think the whole namespace management section of this page needs an overhaul. 
Taking note of this and will follow up in a separate PR

PR #469, opened 2023-08-04
document allowList for github auth

RinkiyaKeDad:
Relevant slack discussion: https://okteto.slack.com/archives/C02T9BKMS4V/p1691047787338819

cc @omnomagonz @ifbyol

PR #588, opened 2024-01-12
document allowlist for github auth

RinkiyaKeDad:
Creating a new PR to fix conflicts in https://github.com/okteto/docs/pull/469

### DIFF
--- a/src/content/reference/helm-chart-values.mdx
+++ b/src/content/reference/helm-chart-values.mdx
@@ -157,6 +157,18 @@ auth:
 
 The `organization` field is optional. Only members of the organization will be allowed to log in into your Okteto instance. An empty `organization` field permits any user to log in.
 
+If you want to provide access to only select users of your organization, you can additionally use the `allowList` key to list the GitHub accounts you want to allow.
+
+```yaml
+auth:
+  github:
+    enabled: true
+    clientId: clientID
+    clientSecret: clientSecret
+    organization: my-org
+    allowList: ["acct1","acct2","acct3"]
+```
+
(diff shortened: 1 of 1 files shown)


src/content/reference/helm-chart-values.mdx
```diff
@@ -151,6 +151,18 @@ auth:
 
 The `organization` field is optional. Only members of the organization will be allowed to log in into your Okteto instance. An empty `organization` field permits any user to log in.
 
+If you want to provide access to only select users of your organization, you can additionally use the `allowList` key to list the GitHub accounts you want to allow.
```
pchico83, line 154: `allowList` is for any user not belonging to the organization, we cannot filter out members of the organization. To do that, you wouldn't define any organization and list all the allowed users in `allowList`

omnomagonz:
```suggestion
You can provide access to select users with the `allowList` key to list the GitHub accounts you want to allow. This feature only works for listing individual GitHub accounts and does not work in conjunction with GitHub organizations.
```

Suggested change based on what I'm understanding here 😄

pchico83:
@RinkiyaKeDad could you update the PR? `allowList` can be used with or without the `organization` key. If there is no `organization`, only these users are allowed to login. Id `organization` is defined, then these users are allowed to login even if they don't belong to your organization. But everyone belonging to the organization is allowed to login

codyjlandstrom:
This PR is getting stale. @RinkiyaKeDad please update, otherwise I'll close and make a ticket next week.

RinkiyaKeDad:
Updated the PR, sorry for the delay!

pchico83 (approved):

PR #593, opened 2024-01-17
Upgrade to Docusaurus v3

huguestennier:
Upgrade to v3 and made some changes following best practices in [the migration guide](https://docusaurus.io/docs/migration/v3)


src/theme/CodeBlock/index.js
```diff
@@ -1,181 +0,0 @@
-/**
```
huguestennier, line 1: We don’t need to do this manually now since Docusaurus provides the copy code button now

codyjlandstrom (approved):
Was able to build it locally first try 😱
Reviewed all pages, links, and 301's. They all are looking good. Thanks, Hugues!

PR #604, opened 2024-01-24
add 1.17 versioned docs

codyjlandstrom:


docusaurus.config.js
```diff
@@ -154,15 +154,20 @@ module.exports = {
...
               label: "1.17",
-              path: "1.17",
+              path: "/",
+              banner: "none",
             },
             1.16: {
               label: "1.16",
-              path: "/",
+              path: "1.16",
               banner: "none",
```
ifbyol, line 171: Should we add the `unmaintained` as we had for `1.15`?

ifbyol (commented):
The build is failing, it seems that some redirection is not right

codyjlandstrom:
> The build is failing, it seems that some redirection is not right

I'm going to submit another PR to fix that quickly.

docusaurus.config.js
codyjlandstrom: Yes, good call 👍

codyjlandstrom:
Just kidding, what I was going to do seems like a larger task. Adding a commit to fix it here.

ifbyol (approved):

PR #608, opened 2024-01-31
Remove redundant overview section

rberrelleza:
This section is already on the first section of https://www.okteto.com/docs/, I think it's redundant here. wdyt?


codyjlandstrom (approved):
Yeah agree this is redundant. Working on some updates for that landing page currently, but this copy doesn't need to be on the overview page

PR #611, opened 2024-02-01
Fix sidebars to allow index pages on sections

rlamana:
This will add page content to sections, removing the need of an "Overview" page.
To do this some styles had to be changed and the sidebar.js file needed to be reorganized. 
In this first example, I did it to the `Okteto Manifest` section, now it doesn't have an "Overview" page. Redirections need to be applied when doing this, as the path changes to `whatever/index` internally, and `whatever` as the canonical URL path.

### Before 

Screenshot 2024-02-01 at 17 19 12

### After 

Screenshot 2024-02-01 at 17 19 02

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -305,6 +305,12 @@ module.exports = {
               "/administration/private-repositories/ssh-key/",
             ],
           },
+          {
+            to: "manifest/index/",
+            from: [
+              "/manifest/overview-manifest/"
+            ]
+          }
         ],
       },
     ],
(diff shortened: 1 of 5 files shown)


rlamana:
Also, the glitch on the sidebar animation has been fixed 😄

codyjlandstrom (approved):
Huge improvement. Two clicks to get to a section was too many. Thanks for doing this!

ifbyol (approved):

docusaurus.config.js
```diff
@@ -305,6 +305,12 @@ module.exports = {
               "/administration/private-repositories/ssh-key/",
             ],
           },
+          {
+            to: "manifest/index/",
```
pchico83, line 309: can I move this to the Netlify redirects config file? any preference between both files?

PR #614, opened 2024-02-02
Add initial docs for Admin Access Tokens

viddo:
Adds some minimal docs for admin tokens, under the _Admin features → Admin dashboard_.

Since we're now introducing this I also reviewed and did some minor updates to other related pages where personal access tokens were previously referenced. E.g. the docs for previews environments and the guides for [Github Actions](http://localhost:8080/docs/preview/github/using-github-actions/) and [Gitlab CI/CD](http://localhost:8080/docs/preview/using-gitlab-cicd/). They had different explanations of _scope_, and was referring _personal_ access tokens. I tried to do minimal changes to make them somewhat congruent, and reference the new admin tokens that I figure we want users to use there instead? I figure the previous recommendation to suffix preview name with user's username should not be needed anymore?

Either way, I think it would make sense to revamp the ["Personal Access Tokens"](http://localhost:8080/docs/core/credentials/personal-access-tokens/) and these guides to both cover and explain the differences between personal and admin access tokens better and when we recommend to use either. But could be done separately, not sure what you have in mind?

While at it, fixed the broken image as can be seen on https://www.okteto.com/docs/preview/using-gitlab-cicd/#step-1-configure-your-okteto-api-token


codyjlandstrom:
> Either way, I think it would make sense to revamp the ["Personal Access Tokens"](http://localhost:8080/docs/core/credentials/personal-access-tokens/) and these guides to both cover and explain the differences between personal and admin access tokens better and when we recommend to use either. But could be done separately, not sure what you have in mind?

Agree here I think it's strange for Personal Access Tokens to have its own page under `Core Concepts` -> `Credentials`. I think we should revamp this separately since it may take some reorganization to make it make sense.

codyjlandstrom (approved):

PR #595, opened 2024-01-19
Add Community Search

puemos:
https://github.com/okteto/docs/assets/13174025/4be17a9d-4018-41bb-bd6a-df956e6a66c0

PR #621, opened 2024-02-07
include community in the search results by using the orbit.love plugin

rberrelleza:


rberrelleza:
@rlamana this is the branch that has the plugin from orbit enabled. Let me know if this is something that we are interested in pursuing, and if you see any blockers/question, and I can put you in touch with them.  I like that this gives people more visibility into community.okteto.com

rberrelleza:
I don't know how stable it is; it's something that orbit recently released. I think it's safe to merge.  It would be cool if we could have a way of measuring if this is helping our users find more answers, but I don't know if your current mixpanel instrumentation can help here. 

Would it be possible to merge it but put it behind a feature flag of some sort, so only say, people from Okteto can see it for now? We can try it for a week or two and then make it the default?

I can put you in touch with the Orbit eng team if you want more information from them.

rlamana:
@rberrelleza we can use env vars to set the docusuarus config file and set this, it would be like putting it behind a feature flag for now, as we test it. I'm not so familiar with netlify, is there a way to set an env var for the preview deploys cc. @huguestennier ?

Btw @rberrelleza, what is the difference between this PR and https://github.com/okteto/docs/pull/595?

rberrelleza:
It's the same, I recreated it so we could get preview environments

PR #635, opened 2024-02-13
Use relative paths to other docs

pchico83:


pchico83:
@codyjlandstrom merging this one to avoid merging hell

PR #638, opened 2024-02-15
Use the recommended version instead of latest

rberrelleza:
Our customers typically copy-paste the doc when configuring a preview environment. 

Having the recommended version for the version of the docs they are looking at should reduce the likelihood of them using the wrong CLI version.

### DIFF
--- a/src/content/preview/github/using-github-actions.mdx
+++ b/src/content/preview/github/using-github-actions.mdx
@@ -42,13 +42,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Context
-        uses: okteto/context@latest
+        uses: okteto/context@2.25.1
         with:
           url: ${{secrets.OKTETO_URL}}
           token: ${{ secrets.OKTETO_TOKEN }}
 
       - name: Deploy preview environment
-        uses: okteto/deploy-preview@latest
+        uses: okteto/deploy-preview@2.25.1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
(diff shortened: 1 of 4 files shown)


rberrelleza:
It would be cool if docusaurus could pull the version from something defined in package.json, or in a version-cliVersion map, but couldn't figure out a way. That could be done in a later PR.

pchico83:
> It would be cool if docusaurus could pull the version from something defined in package.json, or in a version-cliVersion map, but couldn't figure out a way. That could be done in a later PR.

@rlamana could you investigate the best way to have global variables? it would help with different settings like kubernetes and cli versions in our samples

rlamana:
@pchico83 it is using javascript, so it is a matter of just importing a file where the variables are set and use them. Should be simple to do I think. Where do we want to set these variables values? In the package.json? Env vars?

pchico83:
> @pchico83 it is using javascript, so it is a matter of just importing a file where the variables are set and use them. Should be simple to do I think. Where do we want to set these variables values? In the package.json? Env vars?

@rlamana whatever works better for you. If you can add a sample for `KUBERNETES_MIN_VERSION` and `KUBERNETES_MAX_VERSION` I can create independent issues to use variables for other values:
https://github.com/okteto/docs/blob/main/src/content/get-started/cloud-provider-guides/eks.mdx#L33

rberrelleza:
@pchico83 can we merge this in the meantime?

pchico83:
> @pchico83 can we merge this in the meantime?

@rberrelleza check my comments. We recommend using `latest` for github actions, and the tag `stable` for the `okteto/okteto` image in GitLab

rberrelleza:
@pchico83 why do we need two different tags for github actions and gitlab?

pchico83:
> @pchico83 why do we need two different tags for github actions and gitlab? Could we unify to keep things simpler?

@rberrelleza it's legacy. We should standarize in `stable`, which is our cli release channel

PR #640, opened 2024-02-15
Add support for custom variables

rlamana:
New approach to use variables in our docs, and this time versioned:
- We will have a `variables.json` file at the root of our `content/` folder
- In those files that want to use it we will have to import the file and use them as a normal variable. Example:
```mdx
import variables from '../../variables.json';

...
Okteto supports Kubernetes versions {variables.kubernetesMinVersion} through {variables.kubernetesMaxVersion}.
...
```
- Since it lives in the `content/` folder, it will be versioned
- The only downside is that we will always have to reference the file as a relative path to import it.


rlamana:
@pchico83 this is the only way I found to use variables for now. I hope this helps. Let me know if you want to define them based on env vars, or just set the values directly in the docusaurus config file.

pchico83 (approved):
@rlamana this is great. A little more effort and the beginning to render the vars, but it will save a lot of time on every new release :-)

docusaurus.config.js
```diff
@@ -16,6 +16,8 @@ module.exports = {
   onDuplicateRoutes: "warn",
   customFields: {
     image: "https://www.okteto.com/docs/okteto-meta-image.png",
+    kubernetesMinVersion: process.env.KUBERNETES_MIN_VERSION ?? "1.25",
```
rberrelleza, line 19: @rlamana how will this work across future versions? we only want this min/max value to apply to version 1.18, but for version 1.19 we might want this to be 1.26 and 1.29 (for example)

rlamana: I was just trying to figure out that, not sure how to do it just yet. I will keep investigating before merging this

rlamana: I changed the approach, it is now versioned. Read the updated comment and let me know what you think: https://github.com/okteto/docs/pull/640#issue-2137013176

src/content/get-started/cloud-provider-guides/eks.mdx
```diff
@@ -7,6 +7,8 @@ pagination_prev: get-started/overview
 pagination_next: get-started/install-okteto-cli
 ---
 
+import variables from '../../variables.json';
```
rberrelleza, line 10: could we use the format that Hugues used for the [TiersList instead](https://github.com/okteto/docs/tree/main/src/theme/TiersList)? (seems easier to import to me)

rlamana: In the import? No if we want to support this with the versioned docs. Like I mentioned that's the only problem I see with this approach. The import has to be relative since it refers to a different json file in each version, so it has to be relative to the file that is importing it. The `TiersList` is a component that always lives in the "same place", independently of the version and can be referred in absolute terms. As far as I know there are no module "shorthand" for versioned folders, that would be a nice feature.

pchico83: Let's not spend more time on this. The new syntax is perfect for what we want.

PR #641, opened 2024-02-15
New content for "Development environments"

pchico83:
- Add more developer-focused content to the "Development environments" section
- Simplify sidebar


sidebars.js
```diff
@@ -43,23 +43,17 @@ module.exports = {
       type: 'category',
       label: 'Core concepts',
       items: [
-        'core/using-okteto-cli',
-        {
-          type: 'category',
-          label: 'Okteto components',
-          items: [
-            'core/components/development-containers',
```
rberrelleza, line 51: I think you need to add a redirect for this one of the URL okteto.com/docs/core/components/namespace will break

rberrelleza: https://deploy-preview-641--okteto-docs.netlify.app/docs/1.18/core/components/namespace results in a 404

pchico83: It's `namespaces`, not `namespace`. 
This PR doesn't change URLs, it doesn't require any redirects. We don't want to add any redirect at all until the docs revamp is finished. After that, we will probably have a single PR with redirects.
The only redirect is to remove `/docs/guides/tutorials/getting-started-with-pipelines` in favor of `/docs/guides/tutorials/getting-started-with-okteto`

PR #643, opened 2024-02-16
Mark lastest version in dropdown

rlamana:
I think it is confusing to see in the dropdown our current working version and not notify users which one is the latest available version. We can mark it as such like this:

Screenshot 2024-02-16 at 11 06 53

### DIFF
--- a/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.js
+++ b/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.js
@@ -78,9 +78,9 @@ export default function DocsVersionDropdownNavbarItem({
 
       let versionDetails = '';
       if (version.name === 'current') {
-        versionDetails = '(not yet released)';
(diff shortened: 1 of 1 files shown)


pchico83:
@rlamana @codyjlandstrom this is how it looks like in gitlab:
Screenshot 2024-02-16 at 12 39 50
I would also add the arker "not yet released" or something similar

rlamana:
I like it, I will add the "not yet released"

rlamana:
@pchico83 @codyjlandstrom what about this?

Screenshot 2024-02-17 at 00 43 48

codyjlandstrom:
Could we just say "unreleased" to make it a bit shorter? Otherwise looks fine!

rlamana:
@codyjlandstrom Done! 

Screenshot 2024-02-19 at 12 44 33

codyjlandstrom (approved):

PR #652, opened 2024-02-21
mark build metrics as alpha

codyjlandstrom:


pchico83 (approved):

src/content/admin/okteto-insights.mdx
```diff
@@ -108,7 +108,7 @@ Okteto will progressively make more data available through this endpoint. The fo
   - `namespace`: It contains the namespace name.
   - `node_name`: It contains the node name.
 
-### Build Metrics
+### Build Metrics **`ALPHA`**
```
ifbyol, line 112: As now these metrics will be disabled by default and even if you enable `insights` they won't be enabled, I think we should mention how to enable it. For it, apart from enabling Okteto insights feature as we mention at he beginning of this document, they would need to set the helm setting `unsupported.insights.buildMetrics.enabled` to `true`.

Also, would it make sense to include a note or a warning in this section letting them know that as they are `alpha`, these metrics could change in future versions (not all people might understand what `alpha` means

codyjlandstrom: Good feedback! Addressed this in the latest commit

ifbyol (approved):
Looks great! Thanks for adding it!

PR #653, opened 2024-02-22
Add 1.18 versioned docs

codyjlandstrom:


src/content/release-notes.mdx
```diff
@@ -5,6 +5,35 @@ sidebar_label: Release notes
 id: release-notes
 ---
 
+## 1.18.0
+
+27 January 2024
+
+This version is compatible with Kubernetes versions 1.25 to 1.28.
+
+### New Features
```
pchico83, line 14: @codyjlandstrom I would call out the Argo CD installation guide. An alternative is to mentioning it in the release thena message, but I think it's important to let people know about this guide

pchico83 (approved):

src/content/release-notes.mdx
codyjlandstrom: Good call. Will add this now

src/content/release-notes.mdx
```diff
@@ -5,6 +5,36 @@ sidebar_label: Release notes
...
+
+### Bug Fixes
+
+- Hide clear button if namespace search input is empty  <!-- 7667 -->
+- In deploy dialog, fix wrong filter when canceling current repository on SelectRepository  <!--7668 -->
+- Fix filtered namespace view where parent resources were not expanded  <!-- 7669 -->
+- Fix closing create access token dialog from keyboard  <!-- 7671 -->
+- Fix Scale to Zero Period showing wrong value in Previews UI  <!-- 7725 -->
+- Fix wrong "Waking Up" status when switching to a "Sleeping" namespace  <!-- 7817 -->
+- Properly render `code` nodes as inline elements inside a list item  <!-- 7821 -->
```
ifbyol, line 34: Maybe we could add more detail indicating that this is for the customized readmes for external resources and possibly a link to https://www.okteto.com/docs/1.18/manifest/external-resources/#customizing-your-external-resource-ui-elements

ifbyol (commented):
It seems the last commit failed to build but I wasn't able to finde a log error. Could it be related to the issues you were having locally?

src/content/release-notes.mdx
codyjlandstrom: Is that for all 4 of these items @ifbyol

ifbyol: No, only for the one in the line 34 (`Properly render `code` nodes as inline elements inside a list item`)

PR #675, opened 2024-03-20
Added fix of dotnet sample

mnevadom:
Signed-off-by: Mario Nevado <nevado.mario@gmail.com>

Regarding to this fix in the repository: https://github.com/okteto/aspnetcore-getting-started/pull/13

### DIFF
--- a/src/content/development/containers/file-sync/aspnetcore.mdx
+++ b/src/content/development/containers/file-sync/aspnetcore.mdx
@@ -70,7 +70,7 @@ dev:
       - ASPNETCORE_ENVIRONMENT=Development
     remote: 2222
     sync: 
-      - ./:/src
+      - .:/src
 ```
 
 The `hello-world` key matches the name of the hello world Deployment. The meaning of the rest of fields is:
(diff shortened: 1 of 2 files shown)


pchico83:
@mnevadom grep by `/okteto` and you will find a few more changes. 
Also, the prompt shows `app` and I think we should show `src`.
And we need to explain `sync` when explaining the okteto dev section:
```
image: the image used by the development container (built from this [Dockerfile](https://github.com/okteto/aspnetcore-getting-started/blob/main/Dockerfile)).
command: the start command of the development container.
environment: the environment variables added or overwritten in your development container.
remote: the local port to use for SSH communication with your development environment.
```
This is what we say in other samples:
```
sync: the folders that will be synchronized between your local machine and the development container.
```

src/content/development/containers/file-sync/aspnetcore.mdx
```diff
@@ -77,6 +79,7 @@ The `hello-world` key matches the name of the hello world Deployment. The meanin
 - `command`: the start command of the development container.
 - `environment`: the environment variables added or overwritten in your development container.
 - `remote`: the local port to use for SSH communication with your development environment.
+- `sync`: which path is going to be synchorinized from local to container.
```
mnevadom, line 82: @pchico83 Did you mean this section? I can update with the more used text you shared.

pchico83: oh yes! let's use the same text than the other samples

pchico83 (approved):

PR #683, opened 2024-03-22
Document how smart builds works

pchico83:
First draft to document Smart builds. Still pending work with images, but the text is ready for review


codyjlandstrom:
New explanation of smart builds sounds good

src/content/core/build-service.mdx
```diff
@@ -1,26 +1,138 @@
...
+
+```bash
+ i  Using cindy @ okteto.example.com as context
+ i  Building 'api/Dockerfile' in tcp://buildkit.okteto.example.com:443...
+ ...
+ => pushing manifest for registry.okteto.example.com/okteto/api:09f8e312c512e566e82
+ ✓  Image 'registry.okteto.example.com/cindy/api:okteto' successfully pushed
+```
+
+A hash is computed based on the content of the `api` folder, the content of the `Dockerfile` file, and the target `prod`.
```
ifbyol, line 91: Would it make sense to include some build arg in the example to illustrate that? It is mentioned above, but I think adding it to the example might help to understand it

static/img/buildkit.png
rlamana, line 1: Nice diagrams @pchico83 😉  let's meet to talk about these diagrams so I can design something a bit less "manual" 😆

codyjlandstrom: Possible to do this week or early next so we can link to these docs from the release notes?

src/content/core/build-service.mdx
pchico83: Good point, updating!

static/img/buildkit.png
pchico83: Yes, we are working on it today

codyjlandstrom (approved):

PR #695, opened 2024-04-10
A doc explaining remote execution for deploy/destroy commands

pchico83:
This PR adds a dedicate document in the "Core concepts" section explaining remote execution, how it works and its advantages versus local execution


src/content/core/remote-execution.mdx
```diff
@@ -0,0 +1,85 @@
...
+## Ignoring files
+
+Okteto looks for a file named `.oktetoignore` in your local folder. If this file exists, the CLI will exclude any files and directories that match patterns in it.
+This helps to avoid unnecessarily synchronizing large or sensitive files and directories that are not used by the commands defined in the `deploy` section.
+`.oktetoignore` uses the same format as [.dockerignore files](https://docs.docker.com/engine/reference/builder/#dockerignore-file).
+
+## Enable Remote Execution by default
+
+You can also enable Remote Execution by setting the environment variable `OKTETO_FORCE_REMOTE=true`.
+If you define `OKTETO_FORCE_REMOTE=true` as an [Admin Variable](admin/dashboard.mdx#variables), Remote Execution is enable for all your commands.
```
codyjlandstrom, line 79: Does this apply to all commands for that admin user, or for all users across their Okteto instance?

codyjlandstrom (approved):

src/content/core/remote-execution.mdx
pchico83: For all users. Admin variables are retrieved by the okteto cli at the beginning of every command. Defining an admin variable is equivalent to defining that variable on every developer's machine. Would you update the paragraph to make this clearer?

PR #697, opened 2024-04-12
update docusaurus to 3.2.1

codyjlandstrom:
@pchico83, @rlamana This version has some build time improvements that are helpful.
Still seeing some strange processing of redirects, but this version hasn't crashed terminal for me yet.


pchico83 (approved):
@codyjlandstrom that's already a large improvement :-)

PR #709, opened 2024-04-19
[WIP] Update docusaurus to 3.2.1

rlamana:


rlamana:
It seems our problem with version in 3.2.1 is related to this other issue: https://github.com/facebook/docusaurus/issues/10052
I've tried to update to their canary and the fix works in future versions. Let's wait until next release before upgrading docusaurus.

pchico83:
@rlamana thanks for researching what was going on!

PR #723, opened 2024-04-29
Add test to cli reference

maroshii:


src/content/reference/okteto-cli.mdx
```diff
@@ -790,6 +790,26 @@ Displays the current installed version
...
+| _--file_      | (string) | The path to the manifest file                                                                                          | "okteto.yaml" |
+| _--namespace_ | (string) | The namespace to use                                                                                                   | Current Okteto Namespace |
+| _--name_      | (string) | The name of the development environment name to be deployed                                                            | |
+| _--context_   | (string) | The context to use                                                                                                     | Current Okteto Context |
+| _--deploy_    |  (bool)  | Force execution of the commands in the `deploy` section of the okteto manifest                                         | `false` |
+| _--no-cache_  |  (bool)  | Do not use the cache for running the tests                                                                             | `false` |
+| _--var_       |  (list)  | Set a pipeline variable (can be set more than once)                                                                    | |
```
pchico83, line 809: @maroshii we do support `--var` but we don't support `variables` in the test container definition, right?

src/content/reference/okteto-cli.mdx
```diff
@@ -790,6 +790,26 @@ Displays the current installed version
...
+| _--timeout_   | (duration) | The duration to wait for the development environment deployment to complete. Any value should contain a corresponding time unit e.g. 1s, 2m, 3h | 5m0s |
```
pchico83, line 810: @maroshii this is for the deployment of the dev environment? or for the test execution?

pchico83 (approved):

src/content/reference/okteto-cli.mdx
pchico83: Yes, I think that we only support the flag, right? Just asking in case we are missing `variables`, but if it's not supported, it's fine for now. `--var` should say something like `Set an environment variable for the test container (can be set more than once)`

src/content/reference/okteto-cli.mdx
pchico83: Let's document the current behavior. We can improve over it later

pchico83 (approved):

PR #727, opened 2024-05-08
Add documentation to create Catalog Items via CRD

jvc5546:
Documenting solution to a feature request that was closed, due to the solution already existing.

Feature Request for reference: https://okteto.atlassian.net/browse/PROD-152

### DIFF
--- a/src/content/self-hosted/manage/crds.mdx
+++ b/src/content/self-hosted/manage/crds.mdx
@@ -68,3 +68,26 @@ spec:
       key: <field in referenced secret> # key of the secret which contains the access secret key
       name: <secret-name> # name of the secret where the access secret is stored
 ```
+
+## Catalog Items
+
+Catalog items are mainly managed via the [Admin dashboard](admin/dashboard.mdx#catalog-items), but can also be managed via one of the CRDs installed by Okteto: `catalogitems.git.okteto.com`. This last case is useful, for example, if you want to add catalog items manually with a simple `kubectl`, and without the use of the Okteto UI.
+For example, if you want to add a configuration for a catalog item, you will need to create a resource like this:
+
+```yaml title="new-catalog-item.yaml"
+apiVersion: git.okteto.com/v1
+kind: CatalogItem
+metadata:
+  name: my-catalog-item # name of the CatalogItem to be configured
+  namespace: okteto
+spec:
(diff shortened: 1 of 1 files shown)


src/content/self-hosted/manage/crds.mdx
```diff
@@ -68,3 +68,26 @@ spec:
       key: <field in referenced secret> # key of the secret which contains the access secret key
       name: <secret-name> # name of the secret where the access secret is stored
 ```
+
+## Catalog Items
+
+Catalog items are mainly managed via the [Admin dashboard](admin/dashboard.mdx#catalog-items), but can also be managed via one of the CRDs installed by Okteto: `catalogitems.git.okteto.com`. This last case is useful, for example, if you want to add catalog items manually with a simple `kubectl`, and without the use of the Okteto UI.
```
pchico83, line 74: More than `kubectl`, I would mention the scenario to manage your catalog using GitOps

src/content/self-hosted/manage/crds.mdx
```diff
@@ -68,3 +68,26 @@ spec:
...
+For example, if you want to add a configuration for a catalog item, you will need to create a resource like this:
+
+```yaml title="new-catalog-item.yaml"
```
pchico83, line 77:
```suggestion
```yaml title="my-catalog-item.yaml"
```

src/content/self-hosted/manage/crds.mdx
```diff
@@ -68,3 +68,26 @@ spec:
...
+apiVersion: git.okteto.com/v1
+kind: CatalogItem
+metadata:
+  name: my-catalog-item # name of the CatalogItem to be configured
+  namespace: okteto
+spec:
+  branch: "" # default branch fro the CatalogItem. Not required
```
pchico83, line 84: We usually use something like:
```
(Optional). he branch to deploy. Defaults to the default branch of the repo

pchico83: Also, I think it's cleaner to put the sample, and then explain the fields below, as we do here:
https://www.okteto.com/docs/self-hosted/helm-configuration/#volumes

jvc5546: That is a nice approach. I'll change it, thanks!

PR #728, opened 2024-05-08
Add test to preview environment docs

maroshii:
There was no description in the spec about what this should look like. Not sure if this is enough, or what else we want to add and how. @pchico83 @codyjlandstrom can you please review?

### DIFF
--- a/src/content/previews/index.mdx
+++ b/src/content/previews/index.mdx
@@ -16,6 +16,8 @@ Preview environments allow you to include a live preview of your changes on ever
 
 Okteto's preview environments are powered by Kubernetes and easily integrate with your favorite source control and CI/CD provider. They work using the [Okteto manifest](core/okteto-manifest.mdx), so are very simple to get started with.
 
+To unlock the full potential of Okteto Preview Environments, you can also combine it with our [testing infrastructure](reference/okteto-cli.mdx#test) and run integration test against your environments.
+
 - [GitHub](previews/using-github-actions.mdx)
 - [GitLab](previews/using-gitlab-cicd.mdx)
 - Bitbucket Pipelines (coming soon!)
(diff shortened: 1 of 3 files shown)


codyjlandstrom:
Since we're doing an internal only release for the first iteration of Test, I'd like to hold Test docs back until 1.21.
That being said, this looks like a great addition

src/content/previews/using-github-actions.mdx
```diff
@@ -55,6 +55,16 @@ jobs:
           name: pr-${{ github.event.number }}
           timeout: 15m
           scope: global
+
+      # optional
+      # Run integration tests against your newly create preview environment
```
pchico83, line 60: This is what I had in mind too. But we need to add the `test` section to the okteto manifest in the repo

src/content/previews/index.mdx
```diff
@@ -16,6 +16,8 @@ Preview environments allow you to include a live preview of your changes on ever
 
 Okteto's preview environments are powered by Kubernetes and easily integrate with your favorite source control and CI/CD provider. They work using the [Okteto manifest](core/okteto-manifest.mdx), so are very simple to get started with.
 
+To unlock the full potential of Okteto Preview Environments, you can also combine it with our [testing infrastructure](reference/okteto-cli.mdx#test) and run integration test against your environments.
```
pchico83, line 19: I think we need a place to explain the benefits of Okteto Test, @codyjlandstrom where do you think it would be a good place for it?

codyjlandstrom: I think it probably needs it's own section within Preview environments to start. `docs/previews/test`
Another option would be to make it as the same menu level as dev and preview environments for awareness of the feature. We should discuss this as part of the GTM effort

pchico83 (approved):

pchico83 (approved):

PR #764, opened 2024-06-19
Explain why okteto tests with images

pchico83:
This PR adds an explanation about why the Okteto Test is important in your software development cycle.
The PR includes a few images to make it easier to understand the value of Okteto Test.


codyjlandstrom (approved):
Looks fantastic!
My only concern is that it makes the page a bit long, but the explanation is great. Lets definitely add it

pchico83:
> Looks fantastic! My only concern is that it makes the page a bit long, but the explanation is great. Let's definitely add it

@codyjlandstrom I thought the same. I think we will need to rethink the whole test section soon as we add more content

PR #765, opened 2024-06-19
docs: add sc guide for eks (PLAT-572)

jpf-okteto:


pchico83:
@jpf-okteto should we also add it to the 1.21 version?

pchico83 (approved):

PR #770, opened 2024-06-24
Adjust heading levels on `amazon-ecr` page

bgonp:
It looks like using `h1` tags in the content of a page is discouraged, here is a [conversation](https://github.com/facebook/docusaurus/issues/1828) in the Docusaurus repository about it. Docusaurus doesn't inject an ID for first-level headings so we cannot have links directly to these sections. In the [official docs](https://docusaurus.io/docs/markdown-features/toc#markdown-headings) they only talk about `2`, `3` and `4` levels, not mention about `1` AFAIK, but it looks like it generates some issues, e.g. the main title of the page just disappears if there are other `h1`:

|Before|After|
|---|---|
|image|image|

I don't know if we have a policy about how to use headings but it would be great to agree on that and to fix other possible issues like this one that could exist in the project.

(The preview won't work until https://github.com/okteto/docs/pull/769 is merged)

### DIFF
--- a/src/content/admin/registry-credentials/amazon-ecr.mdx
+++ b/src/content/admin/registry-credentials/amazon-ecr.mdx
@@ -11,7 +11,7 @@ It's recommended that you have the [AWS CLI installed](https://docs.aws.amazon.c
 
 ECR credentials can be configured with either static credentials belonging to an IAM user or using OIDC federation to assume an IAM Role via Web Identity.
 
-# Using IAM User credentials
+## Using IAM User credentials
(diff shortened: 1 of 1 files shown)


codyjlandstrom:
Yeah it can be confusing, but we should stick to only using H2, H3, etc. in our docs and only H1 for the Docusaurus generated header for SEO purposes. I believe the latest google search algos penalizes you if more than one H1 appears

codyjlandstrom (approved):

PR #779, opened 2024-07-10
Create archived release notes page

codyjlandstrom:
Create separate page for all archived release notes. This is to allow the Algolia scraper to index the full release notes page - previously it was too long to capture all content.


rberrelleza:
@codyjlandstrom we already have https://www.okteto.com/docs/archives/, where each legacy release it's linked to the release notes of that release. Do you think we also need the link in the footer? Seems a bit redundant to me to have them on two different places. 

image

codyjlandstrom:
> My personal feedback is to remove `src/content/archived-release-notes.mdx` and leave the archive page as it was before this PR. wdyt?

@rberrelleza Do you think we still need to have the release notes prior to 1.10 published some place?

I feel like we should keep those just in case, but not sure where makes the most sense without adding another page.

rberrelleza:
@codyjlandstrom ah I see what you mean. I didn't realize that in the archive page we only showed up to 1.11. Your original suggestion makes sense in that case!

rberrelleza (approved):

PR #782, opened 2024-07-11
Clarify okteto test --no-cache option

viddo:
Basically provide some more info to _why_ one might want to disable the caching.


pchico83 (approved):
@viddo could you also update this file? that way, the change also apply to current docs version:
https://github.com/okteto/docs/blob/main/versioned_docs/version-1.22/reference/okteto-cli.mdx

viddo:
> @viddo could you also update this file? that way, the change also apply to current docs version: https://github.com/okteto/docs/blob/main/versioned_docs/version-1.22/reference/okteto-cli.mdx

Ah my bad! Didn't realize we just published the new version. Also noticed we have `okteto test` docs back to v1.21 so applied for both v1.21 + v1.22, considering the docs clarification is independent from the _chart_/CLI version.

ifbyol (approved):

PR #784, opened 2024-07-15
add example to fix role-binding issue with cluster-admin role

andreafalzetti:
I am proposing to add a paragraph going a bit deeper showing an example on how to deal with potential issues caused by the removal of this feature

### DIFF
--- a/src/content/self-hosted/manage/upgrade.mdx
+++ b/src/content/self-hosted/manage/upgrade.mdx
@@ -94,6 +94,30 @@ Old settings are still considered and have preference, but their support will be
 
 We have also made another relevant change to consider when migrating to `1.22`. We have stopped creating an automatic role binding to any custom service account created on namespaces managed by Okteto to the cluster role specified on `serviceAccounts.roleBindings.namespaces` (`clusterRole` in previous versions). If you create any custom service account as part of your application, make sure you also create any role or role binding needed.
 
+If you experience issues with permissions in this regard, we recommend creating an explicit role and role-binding, but as an alternative solution you could also manually just create the role-binding to the `cluster-admin` role, using the following approach:
+
+1. Create the `rb.yml.tpl` file with the following content:
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: <service-account-name>
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+- kind: ServiceAccount
+  name: <service-account-name>
+  namespace: ${OKTETO_NAMESPACE}
+```
+2. Replace `<service-account-name>` with the name of the service account you want to bind to the `cluster-admin` role.
+3. In the Okteto Manifest deploy commands, use  `envsubst` to replace the `${OKTETO_NAMESPACE}` variable:
+```yaml
+deploy:
+  - envsubst < k8s/rb.yml.tpl > k8s/rb.yml
+  - kubectl apply -f k8s
+```
 
 ## Upgrading to Okteto 1.20.x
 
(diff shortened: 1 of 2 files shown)


src/content/self-hosted/manage/upgrade.mdx
```diff
@@ -94,6 +94,30 @@ Old settings are still considered and have preference, but their support will be
...
+subjects:
+- kind: ServiceAccount
+  name: <service-account-name>
+  namespace: ${OKTETO_NAMESPACE}
+```
+2. Replace `<service-account-name>` with the name of the service account you want to bind to the `cluster-admin` role.
+3. In the Okteto Manifest deploy commands, use  `envsubst` to replace the `${OKTETO_NAMESPACE}` variable:
+```yaml
+deploy:
+  - envsubst < k8s/rb.yml.tpl > k8s/rb.yml
```
pchico83, line 118: These two files should be in different folders. Otherwise, `kubectl apply -f k8s` will consider both files

andreafalzetti: Are you sure? I assumed the `.tpl` extension would make it skip, I also tried and it seems to skip it.

I guess each team/user has it's preference on how to organize their files and suggesting a completelly different folder might be cleaner but then it focrces to have the additional dir.

pchico83: Oh yes, if `kubectl` skips `*.tpl` extensions, I would keep it as it is

pchico83 (approved):

src/content/self-hosted/manage/upgrade.mdx
```diff
@@ -94,6 +94,30 @@ Old settings are still considered and have preference, but their support will be
 
 We have also made another relevant change to consider when migrating to `1.22`. We have stopped creating an automatic role binding to any custom service account created on namespaces managed by Okteto to the cluster role specified on `serviceAccounts.roleBindings.namespaces` (`clusterRole` in previous versions). If you create any custom service account as part of your application, make sure you also create any role or role binding needed.
 
+If you experience issues with permissions in this regard, we recommend creating an explicit role and role-binding, but as an alternative solution you could also manually just create the role-binding to the `cluster-admin` role, using the following approach:
```
codyjlandstrom, line 97:
```suggestion
If you experience issues with permissions in this regard, we recommend creating an explicit role and role-binding, but as an alternative solution you can also manually create the role-binding to the `cluster-admin` role, using the following approach:
```

versioned_docs/version-1.22/self-hosted/manage/upgrade.mdx
```diff
@@ -94,6 +94,31 @@ Old settings are still considered and have preference, but their support will be
 
 We have also made another relevant change to consider when migrating to `1.22`. We have stopped creating an automatic role binding to any custom service account created on namespaces managed by Okteto to the cluster role specified on `serviceAccounts.roleBindings.namespaces` (`clusterRole` in previous versions). If you create any custom service account as part of your application, make sure you also create any role or role binding needed.
 
+If you experience issues with permissions in this regard, we recommend creating an explicit role and role-binding, but as an alternative solution you could also manually just create the role-binding to the `cluster-admin` role, using the following approach:
```
codyjlandstrom, line 97:
```suggestion
If you experience issues with permissions in this regard, we recommend creating an explicit role and role-binding, but as an alternative solution you can also manually create the role-binding to the `cluster-admin` role, using the following approach:
```

codyjlandstrom (approved):

src/content/self-hosted/manage/upgrade.mdx
andreafalzetti: I double checked this by having a directory with a `k8s.yml.tpl` inside it and run apply against it, you get the following error:

```
[recognized file extensions are [.json .yaml .yml]](error: error reading [k8s]: recognized file extensions are [.json .yaml .yml])
```

PR #785, opened 2024-07-17
update java guide to be more concise

codyjlandstrom:
☕️


pchico83 (approved):
@codyjlandstrom all the samples for dev containers with file sync use the same wording and explanations, should we apply these changes to all of them?

codyjlandstrom:
> @codyjlandstrom all the samples for dev containers with file sync use the same wording and explanations, should we apply these changes to all of them?

Totally agree. I created a task in the docs project to complete this 👍

PR #801, opened 2024-08-06
make it more evident that we are talking about Kubernetes services 

rberrelleza:
make it more evident that annotation examples we are showing in [are talking about Kubernetes services ](https://www.okteto.com/docs/core/endpoints/automatic-ssl/#auto-create-ingresses-from-services) are about Kubernetes services and not Okteto manifests

### DIFF
--- a/src/content/core/endpoints/automatic-ssl.mdx
+++ b/src/content/core/endpoints/automatic-ssl.mdx
@@ -15,15 +15,7 @@ If you've configured Endpoints but they aren't appearing, please check our [FAQ 
 
 ## Auto create ingresses from services
 
-Okteto can automatically create an SSL endpoint for your deployments. In order to take advantage of this feature, add the annotation below to your service's manifest:
-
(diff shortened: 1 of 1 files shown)


src/content/core/endpoints/automatic-ssl.mdx
```diff
@@ -15,15 +15,7 @@ If you've configured Endpoints but they aren't appearing, please check our [FAQ
 
 ## Auto create ingresses from services
 
-Okteto can automatically create an SSL endpoint for your deployments. In order to take advantage of this feature, add the annotation below to your service's manifest:
-
-```yaml
-metadata:
-  annotations:
-    dev.okteto.com/auto-ingress: "true"
-```
```
codyjlandstrom, line 24: Are we sure we want to delete these examples? I guess their taking up a lot of room and not helping too much

codyjlandstrom (approved):

PR #806, opened 2024-08-09
Create 1.23 Release Version

codyjlandstrom:


ifbyol (approved):
Approving it to unblock the merge, but we should wait until the chart release is done to merge it. I asked @maroshii to announce it, so you are aware when it happens

PR #810, opened 2024-08-16
Release notes for 1.23.2

maroshii:


ifbyol (approved):

src/content/release-notes.mdx
```diff
@@ -5,6 +5,16 @@ sidebar_label: Release notes
...
 
+## 1.23.2
+
+16 August 2024
+
+This version is compatible with Kubernetes versions 1.26 to 1.30
+
+### Bug Fixes
+
+- Fix deploys for repositories with a trailing slash
```
codyjlandstrom, line 16: Should we clarify that this is for the catalog?

codyjlandstrom (approved):

src/content/release-notes.mdx
ifbyol: It would happen also if they deploy it from the CLI for example. It was detected with the catalog, but it might happen from other places

PR #813, opened 2024-08-22
fix OKTETO_REGISTRY_URL var

andreafalzetti:
I think this is an old typo we never noticed because `OKTETO_REGISTRY` does not exist, nor appears to have existed in the past.

I've checked [Okteto CLI 2.0.0](https://github.com/okteto/okteto/blob/release-2.0/pkg/model/const.go#L202) and it was still called `OKTETO_REGISTRY_URL` like today: https://github.com/okteto/okteto/blob/master/pkg/model/const.go#L150

To validate run `okteto deploy` with this manifest

```
deploy:
  - env | grep OKTETO_REG
```

### DIFF
--- a/src/content/core/container-registry.mdx
+++ b/src/content/core/container-registry.mdx
@@ -68,7 +68,7 @@ This command will show you information about the current okteto context in use:
 ```
 
 You can retrieve your username, password (token), and the domain for your registry from the output.
-You can also use the environment variables `OKTETO_USERNAME`, `OKTETO_TOKEN`, `OKTETO_NAMESPACE` and `OKTETO_REGISTRY` if your script is running as part of the `deploy`commands of your Okteto Manifest.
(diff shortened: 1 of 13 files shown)


src/content/core/container-registry.mdx
```diff
@@ -68,7 +68,7 @@ This command will show you information about the current okteto context in use:
 ```
 
 You can retrieve your username, password (token), and the domain for your registry from the output.
-You can also use the environment variables `OKTETO_USERNAME`, `OKTETO_TOKEN`, `OKTETO_NAMESPACE` and `OKTETO_REGISTRY` if your script is running as part of the `deploy`commands of your Okteto Manifest.
```
ifbyol, line 71: Should we change it also in the `helm` code sample below? We are referring also to `OKTETO_REGISTRY` there

andreafalzetti: ooh thanks, I had missed it. Updated now

ifbyol (approved):
Good catch!! 👏

PR #817, opened 2024-08-30
Initial docs for multi user

pchico83:
We will iterate the content more and add screenshots once the feature is ready, but sending an initial PR.
One of the reasons is to have a final link to document how to retrieve a User Token using `kubectl`, which will be pointed from the UI

### DIFF
--- a/src/content/self-hosted/install/auth/token.mdx
+++ b/src/content/self-hosted/install/auth/token.mdx
@@ -9,11 +9,26 @@ id: token
 
 Okteto defaults to token-based authentication to simplify the installation and configuration of Okteto self-hosted.
 
-This enables you to use a token for a single user, login to the Okteto UI, and start exploring Okteto's features and functionality much faster than configuring other authentication methods.
-Other authentication methods are more time-consuming and cumbersome to configure.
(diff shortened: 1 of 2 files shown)


src/content/self-hosted/install/auth/token.mdx
```diff
@@ -9,11 +9,26 @@ id: token
...
+
+Okteto generates "Invitation Links" in Token Auth mode to allow other users to log in to Okteto.
+When a user logs in to Okteto via the Invitation Link, a "Welcome screen" is presented showing the user's token.
+This token can be used to log in from the Okteto UI or using the Okteto CLI.
+
+The "Welcome screen" prompts the user to store this token securely.
+In the event of losing this token, the cluster admin can retrieve the user's token by executing the following command:
+
+```bash
+kubectl ...
```
pchico83, line 33: @ifbyol what would be this command?

ifbyol: @pchico83 the command would be ` kubectl get sa -l=dev.okteto.com/external-id=<username> -o=jsonpath='{.items[0].metadata.labels.dev\.okteto\.com/token}'`

pchico83: thanks, updated!

codyjlandstrom (approved):

PR #827, opened 2024-09-11
A general pass adapting docs to CLI 3.0

pchico83:
First pass with CLI 3.0

I also fixed some docs unrelated to CLI 3.0 that I found while reading all docs


src/content/reference/faqs.mdx
```diff
@@ -53,36 +53,7 @@ Please see our [ArgoCD Configuration Guide](self-hosted/manage/argocd.mdx) for o
 
 ## How to use private images?
 
-In order to use your private registry credentials you must configure a [pull secret](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/) and use it in your Kubernetes manifests.
```
pchico83, line 56: This is a very old explanation, we now have Registry Credentials supported at the Okteto Platform level

codyjlandstrom (approved):

PR #830, opened 2024-09-11
Release notes for 1.24.2

maroshii:


pchico83 (approved):

codyjlandstrom (approved):

src/content/release-notes.mdx
```diff
@@ -5,6 +5,17 @@ sidebar_label: Release notes
 id: release-notes
 ---
 
+## 1.24.2
+
+9 September 2024
+
+This version is compatible with Kubernetes versions 1.27 to 1.30
```
codyjlandstrom, line 12: @pchico83 We're ok mentioning vulns here?

pchico83: Yes, it's the only things in the patch release. For major release I don't think it provides value (unless we do a CVE report), but patch releases is different

PR #826, opened 2024-09-11
Add Resource Manager docs page

codyjlandstrom:

PR #834, opened 2024-09-13
Add `resourceManager` helm config

andreafalzetti:
Needs: https://github.com/okteto/docs/pull/826

This PR updates the Helm configuration docs, adding the `resourceManager`


src/content/self-hosted/helm-configuration.mdx
```diff
@@ -485,6 +485,35 @@ regcredsManager:
...
 ```
 
+### resourceManager
+
+Use this section to configure the [Resource Manager](admin/resource-manager.mdx)
+
+- `enabled`: When set to `true`, the Resource Manager is installed. When Resource Manager is installed, Okteto creates a CronJob responsible for computing the cpu and memory recommendations for your cluster.
+- `schedule`: Defines the frequency at which resource recommendations are calculated, using the cron syntax.
+- `deletePeriod`: Specifies the retention period for the recommendations. If a service is no longer deployed in your cluster, its recommendation will be deleted after the specified deletePeriod.
+- `recommendations.weight`: A floating-point number between 0 and 1 used to stabilize CPU and memory recommendation calculations. A higher value (closer to 1) gives more weight to historical resource usage trends, making the Resource Manager less responsive to recent changes or spikes. A lower value (closer to 0) makes it more reactive to recent usage patterns, including spikes. The default value is `0.9`, which makes the Resource Manager conservative, favoring existing trends and attempting to ignore minor fluctuations.
```
andreafalzetti, line 495: @codyjlandstrom can you proof read the initial draft, and tell me your thoughts?

With @pchico83 we are not 100% sure on the naming for `weight` and `correction`. 

1. Is it clear what they are for?
2. do you have any better names?

🙏

pchico83 (commented):
@andreafalzetti Great work explaining the different fields.
I would mention that by default, the RM is in "manual" mode and it only computes recommendations without applying it, and they can be seen in the namespace/metrics tab. And that to apply the recommendations, you need to set it to "automatic" from the admin view.
I would also mention the two annotations to customize the key and to ignore the RM for a particular pod

src/content/self-hosted/helm-configuration.mdx
```diff
@@ -485,6 +485,35 @@ regcredsManager:
       memory: 800Mi
 ```
 
+### resourceManager
+
+Use this section to configure the [Resource Manager](admin/resource-manager.mdx)
+
+- `enabled`: When set to `true`, the Resource Manager is installed. When Resource Manager is installed, Okteto creates a CronJob responsible for computing the cpu and memory recommendations for your cluster.
+- `schedule`: Defines the frequency at which resource recommendations are calculated, using the cron syntax.
+- `deletePeriod`: Specifies the retention period for the recommendations. If a service is no longer deployed in your cluster, its recommendation will be deleted after the specified deletePeriod.
```
andreafalzetti, line 494:
```suggestion
- `deletePeriodDays`: Specifies the retention period for the recommendations. If a service is no longer deployed in your cluster, its recommendation will be deleted after the specified deletePeriod.
```

src/content/self-hosted/helm-configuration.mdx
codyjlandstrom: The initial draft sounds great, @andreafalzetti.

1. I actually like these names and think they are pretty clear
2. `correction` could potentially be "compensation" but I think both are easily understandable. Users configuring these will have to refer to the documentation to understand the values regardless, so I don't think there is much difference in the name

src/content/self-hosted/helm-configuration.mdx
```diff
@@ -485,6 +485,35 @@ regcredsManager:
       memory: 800Mi
 ```
 
+### resourceManager
+
+Use this section to configure the [Resource Manager](admin/resource-manager.mdx)
+
+- `enabled`: When set to `true`, the Resource Manager is installed. When Resource Manager is installed, Okteto creates a CronJob responsible for computing the cpu and memory recommendations for your cluster.
```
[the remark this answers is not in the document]
andreafalzetti: > Instead of "for your cluster", I would say "for your pods"

@pchico83 I've updated the copy. 

> I would say "for your pods" and I would be explicit about that the recommendation is for cpu/memory requests based on real time utilization using the Metrics API

If we document that in the dedicated page for Resource Manager, we can keep this section only dedicated to the actual helm config

src/content/self-hosted/helm-configuration.mdx
```diff
@@ -483,6 +483,39 @@ regcredsManager:
       memory: 800Mi
 ```
 
+### resourceManager
+
+Use this section to configure the [Resource Manager](admin/resource-manager.mdx)
+
+- `enabled`: When set to `true`, the Resource Manager is installed. When Resource Manager is installed, Okteto creates a CronJob responsible for computing the cpu and memory recommendations for your pods. It defaults to `false`.
```
pchico83, line 490: Minor: could you sort the fields in alphabetical order?

pchico83 (approved):

src/content/self-hosted/helm-configuration.mdx
andreafalzetti: @pchico83 shall we sort them also in `values.yaml`?

PR #838, opened 2024-09-18
Clarify the events that wake a sleeping resource

codyjlandstrom:


src/content/admin/cleanup.mdx
```diff
@@ -20,15 +20,26 @@ Learn how to configure this in the [configuration steps](#configuration-steps-).
...
 
-Okteto's UI will notify you when there are sleeping resources in your namespace. Press the **Wake all** button to activate all sleeping resources at once.
+The following events will reset the inactivity counter, preventing the resources from entering a sleep state:
+
+- Deploying a Kubernetes Deployment or StatefulSet with `kubectl` or similar tools
+- Running `okteto deploy`
+- Running `okteto up`
+- Maintaining an active `okteto up` session. If the `upSessionByLastSyncedFile` setting is enabled, this only counts if files are actively being synced (based on the last synced file)
+
+## Wake Sleeping Resources
```
andreafalzetti, line 38: I know it's not part of this change, but we have a sample screenshot for how to wake resources, but not for how to sleep them with the UI. Add perhaps?

andreafalzetti (approved):
LGTM, left a non-blocking comment

src/content/admin/cleanup.mdx
```diff
@@ -20,15 +20,26 @@ Learn how to configure this in the [configuration steps](#configuration-steps-).
...
-- Launched a development container with `okteto up`.
-- Deploy the latest version of your code with `okteto deploy`.
+- Upgrade or redeploy via the UI or the command line
+- Launched a development container with `okteto up`
+- Deploy the latest version of your code with `okteto deploy`
 
 Service accounts, config maps, secrets, or volumes you create will be unaffected by this operation.
 
-## Wake sleeping resources
+### Events That Count as Activity to Reset the Sleep Counter
```
pchico83, line 29: This events, and the previous one below `Okteto considers an application or resources inactive if a user hasn't performed any of the following tasks during the `inactivity` period` are the same.
Each time there is any this events, we considered there is activity, and we reset the sleep counter.

src/content/admin/cleanup.mdx
codyjlandstrom: Good callout. We're also missing information on how to manually sleep a resource from the namespace list view. Will add.

src/content/admin/cleanup.mdx
codyjlandstrom: I called them out separately since the content was slightly different. I also feel like a user searching for one thing may not necessarily read the whole section. Happy to consolidate though if you feel like it's too much

pchico83: I think they are the same. We are explaining what keeps a namespace awaken. It's the same events.

codyjlandstrom: Ok, I shifted the content a bit to support this. Let me know what you think

codyjlandstrom: @pchico83 The latest changes should fix the duplication. Let me know if that looks alright to you

pchico83 (approved):

PR #839, opened 2024-09-18
docs scenario sharing vars deploy to destroy

andreafalzetti:
With this PR we aim to document the behaviour that the Okteto CLI implements, about propagating the variables of `deploy` to the `destroy`.

I am using the term `command item` to refer to the specific line/command/array item, but it overlaps with the "CLI command" such as deploy/destroy, so if someone has better copy, that is less confusing, please share it :)

### DIFF
--- a/src/content/core/okteto-variables.mdx
+++ b/src/content/core/okteto-variables.mdx
@@ -92,7 +92,35 @@ deploy:
 
 ### Sharing variables between commands
 
-As mentioned in the section [Scope](#scope), variables defined in a specific command item are only available to that item by default. In the following sections, there are two ways to share values between steps.
+As mentioned in the section [Scope](#scope), variables defined in each command of your deploy/destroy/test sections only available to that item by default.
+
+For example, here we have two command items, in which the variable `MY_VAR` is **not propagated**:
+
+```
+deploy:
+  - export MY_VAR=hello
+  - echo $MY_VAR # -> '' (empty)
+```
+
+In the following sections, you will see different ways to share values between steps or CLI commands.
+
+#### Sharing Variables Between Deploy and Destroy
+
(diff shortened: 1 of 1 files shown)


src/content/core/okteto-variables.mdx
```diff
@@ -92,7 +92,35 @@ deploy:
...
+deploy:
+  - export MY_VAR=hello
+  - echo $MY_VAR # -> '' (empty)
+```
+
+In the following sections, you will see different ways to share values between steps or CLI commands.
+
+#### Sharing Variables Between Deploy and Destroy
+
+Okteto automatically shares variables passed with the `--var` flag during the `deploy` command with the `destroy` command. This ensures consistent configuration across both deployment and teardown processes without the need to redefine variables.
```
pchico83, line 109: Does it apply to `OKTETO_ENV`? If yes, I would mention it

src/content/core/okteto-variables.mdx
```diff
@@ -92,7 +92,35 @@ deploy:
 
 ### Sharing variables between commands
 
-As mentioned in the section [Scope](#scope), variables defined in a specific command item are only available to that item by default. In the following sections, there are two ways to share values between steps.
+As mentioned in the section [Scope](#scope), variables defined in a specific command item are only available to that item by default.
```
pchico83, line 95: I would use something like "each command of your deploy section", and mention destroy/test sections too

codyjlandstrom (approved):

src/content/core/okteto-variables.mdx
andreafalzetti: changed to:

```diff
-... in a specific command item are only available to that item by default.
+... in each command of your deploy/destroy/test sections only available to that item by default.
```

WDYT?

PR #857, opened 2024-09-29
Add dependencies to manifest documentation. Add namespace to last versioned docs.

### LINKED ISSUE #420: Document manifest property `dependencies.<item>.namespace`
As per the title!

seanson:
Resolves #420 
This adds documentation to the extended notation for dependencies in the Okteto manifest.

The contributing docs mentions that only current documentation should be modified but I am not sure how that works with documentation changes that are both present in previous versions and about to change with the 3.x release, so I have added an extra change to the 1.24 version documentation.

If this is not ideal then I'm glad to amend the commit.

### DIFF
--- a/src/content/reference/okteto-manifest.mdx
+++ b/src/content/reference/okteto-manifest.mdx
@@ -96,6 +96,8 @@ Follow this document for a list of [environment variables](reference/okteto-cli.
 
 Use `okteto deploy --dependencies` to force the redeployment of your dependencies.
 
+### dependencies - extended notation (object, optional)
+
 There is also an extended notation to configure how to deploy your dependency:
 
 ```yaml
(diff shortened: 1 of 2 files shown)


src/content/reference/okteto-manifest.mdx
```diff
@@ -96,6 +96,8 @@ Follow this document for a list of [environment variables](reference/okteto-cli.
 
 Use `okteto deploy --dependencies` to force the redeployment of your dependencies.
 
+### dependencies - extended notation (object, optional)
```
pchico83, line 99: Let's remove this line. That way everything is index under `dependencies`

src/content/reference/okteto-manifest.mdx
```diff
@@ -111,22 +113,45 @@ dependencies:
     timeout: 15m
 ```
 
-Or:
+#### repository (string)
```
pchico83, line 116: I would use the same format that we use in the [build section](https://www.okteto.com/docs/reference/okteto-manifest/#build-object-optional). We put the sample first, and then a list of items explaining each field

src/content/reference/okteto-manifest.mdx
seanson: Is there a good way we can communicate that this property has a simple and an extended notation then? I like the idea of the formatting being consistent but need a way as a consumer to know that there are two explicit versions. (Sidenote, is there a YAML schema for the Okteto file somewhere?)

pchico83: We don't have a YAML schema for the Okteto Manifest yet :disappointed_relieved:
Are you familiar with any documentation where they document different field types?
I have been looking for inspiration, but didn't find anything great

PR #861, opened 2024-10-01
Refactor The Okteto CLI installation instructions

pchico83:
I have been experimenting with the location of the Okteto CLI installation instructions. In the current layout, it's inside the "Getting Started for Platform Engineers" section, which looks weird.

This PR moves the Okteto installation guide to the top of the "Get Started" section. This can also be useful if we start implementing Okteto CLI commands to help with the Okteto Chart installation.

Other changes:
- Remove the Okteto CLI configuration step from the CLI installation guide
- Explain how to configure the CLI as part of the Chart installation guide
- Use the Okteto CLI installation guide as a single point of true and link to it from the Developer's Quickstart

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -12,6 +12,7 @@ module.exports = {
       label: 'Get Started',
       link: { type: 'doc', id: 'index' },
       items: [
+        "get-started/install-okteto-cli",
         {
           type: "category",
           label: "For Platform Engineers",
@@ -29,7 +30,6 @@ module.exports = {
                 "get-started/install/openshift"
               ]
             },
-            "get-started/install-okteto-cli",
             {
               type: "category",
               label: "Deploy your app",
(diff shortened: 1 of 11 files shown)


rlamana:
Agree that it should be at that first level, but wondering if we should place it below the other 2, that way for platform engineers > Install Okteto (the platform) is the first option, rather than installing the CLI. Just a suggestion, I think this is OK too.

codyjlandstrom (approved):

pchico83:
> Agree that it should be at that first level, but wondering if we should place it below the other 2, that way for platform engineers > Install Okteto (the platform) is the first option, rather than installing the CLI. Just a suggestion, I think this is OK too.

The Okteto CLI install is needed to complete the platform engineering guides, because the "deploy your first app" is part of that section. Also, in the future, the CLI might help with the chart installation

PR #862, opened 2024-10-01
Fix inline code blocks to look vertically balanced

viddo:
The base paragraph text is actually 18px, but for whatever reason the code block was at some point changed 16px. This causes the text to look misaligned and wonky when there's a mix of regular and inline `code` texts present on same lines.

It appears the vertical align was set to _middle_ to [fix alignment in the default Docusaurus styles](https://github.com/facebookincubator/infima/pull/113) (aka [Infirma](https://docusaurus.io/docs/next/styling-layout#styling-your-site-with-infima)), but that clashes with our custom body and code fonts.

Fixed by setting 18px as default size also for inline code blocks, and resetting the `vertical-align` to _baseline_ instead.

While at it, also removing unnecessary props like `border-radius` and `border` that are already set through the underlying default styles.

## Testing done

Reviewing changes in all evergreen browsers, comparing with main, e.g.

| Before | After |
| -- | -- |
| Screenshot 2024-10-01 at 15 55 36 | Screenshot 2024-10-01 at 15 55 31 |
| Screenshot 2024-10-01 at 15 56 04 | Screenshot 2024-10-01 at 15 56 19 |
| Screenshot 2024-10-01 at 15 57 56 | Screenshot 2024-10-01 at 15 58 05 |

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -51,7 +51,7 @@
 
   --ifm-font-family-monospace: #{$code-font};
   --ifm-code-background: #{$light-blue};
-  --ifm-code-font-size: 16px;
+  --ifm-code-font-size: 18px;
 
   --ifm-tabs-padding-vertical: 12px;
 }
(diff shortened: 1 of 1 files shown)


bgonp (approved):

codyjlandstrom (approved):
Looks so much better! Thank you

PR #864, opened 2024-10-01
Add 1.25 release notes

codyjlandstrom:


src/content/release-notes.mdx
```diff
@@ -5,6 +5,43 @@ sidebar_label: Release notes
...
+## 1.25.0
+
+7 October 2024
+
+This versions is compatible with Kubernetes versions 1.27 to 1.30
+
+### New Features
+
+- [Announcing Okteto CLI 3.0](https://www.okteto.com/blog/cli-three-release/) - upgrading to chart release 1.25 requires a CLI upgrade to 3.0
+- Announcing Cloud Credentials, a central location to manage your authentication with AWS and GCP  <!-- 8550 -->
```
ifbyol, line 17: Should we specify that is for remote operations (`deploy`, `destroy` and `test`)? Also a link to [here](https://www.okteto.com/docs/1.25/admin/cloud-credentials/) might help for docs disceverability?

src/content/release-notes.mdx
```diff
@@ -5,6 +5,43 @@ sidebar_label: Release notes
...
+- Enabling Okteto Insights for all users of Okteto <!-- 8627 -->
```
ifbyol, line 18: I would specify this is only for SaaS environments

src/content/release-notes.mdx
```diff
@@ -5,6 +5,43 @@ sidebar_label: Release notes
...
+- Added configurable buildkit network mode  <!-- 8486 -->
```
ifbyol, line 19: Should we link to [here](https://www.okteto.com/docs/1.25/self-hosted/helm-configuration/#buildkit)?

src/content/release-notes.mdx
```diff
@@ -5,6 +5,43 @@ sidebar_label: Release notes
...
+- Init dynamic config  <!-- https://github.com/okteto/app/pull/8475 -->
```
ifbyol, line 20: This is something internal. Not sure we should document it. Maybe we should approach it as Okteto installing a new CRD in your cluster? cc @pchico83

src/content/release-notes.mdx
```diff
@@ -5,6 +5,43 @@ sidebar_label: Release notes
...
+- Added CLI, Chart, and Okteto App version to our API headers  <!-- 8545 -->
+- Support to enable/disable resource quotas for individual Namespaces  <!-- 8604 -->
```
ifbyol, line 22: This change is related to allow to enable only specific resource quotas instead of all of them for namespaces. I mean, before, if you enabled quotas, you would restrict all the resources of the quotas (pods, services, secrets, configmaps, etc..)

Now, with this change, you can enable the quotas and restrict only one or a subset of resources, allowing to create any amount of the other resources. For example, you could restrict only the number of pods.

I think it could be relevant to highlight here that we have changed the default values. If there was a user that enabled only the quotas, they were using the previous default values that now are `-1`. 

We could also link to here: https://www.okteto.com/docs/1.25/self-hosted/helm-configuration/#quotas

src/content/release-notes.mdx
```diff
@@ -5,6 +5,43 @@ sidebar_label: Release notes
...
+- Allow override of installer security context  <!-- https://github.com/okteto/app/pull/8608 -->
```
ifbyol, line 23: We should document this field in the docs [here](https://www.okteto.com/docs/1.25/self-hosted/helm-configuration/#installer) and probably add an example on the code snippet there. It is not easy to know how to do it without docs cc @maroshii

src/content/release-notes.mdx
```diff
@@ -5,6 +5,43 @@ sidebar_label: Release notes
...
+
+- Show more actionable feedback upon "failed to deploy okteto pipeline"…  <!-- 8521 -->
+- You can now list your Github App installations under Settings → Integrations  <!-- 8463 -->
+- Implemented a number of UI updates to ensure your Okteto experience stays smooth <!-- 8600, 8623, 8534, 8626 -->
+
+### Bug Fixes
+
+- Fixed wrong data parsing error from unknown git urls  <!-- 8504 -->
+- Fixed collapsed content not being accessible   <!-- 8522 -->
+- Replaced `Input` with `FormField` in the login view  <!-- 8510 -->
```
ifbyol, line 35: Not sure this is relevant for users. The one about the input seems more internal.

The first is hard to know if it is affecting something customer facing or not. It doesn't specify where this content wasn't being accessible. Is that relevant for users?

src/content/release-notes.mdx
pchico83: This was already shipped in the previous version, right?

src/content/release-notes.mdx
pchico83: Agree

src/content/release-notes.mdx
```diff
@@ -5,6 +5,43 @@ sidebar_label: Release notes
 id: release-notes
 ---
 
+## 1.25.0
+
+7 October 2024
+
+This versions is compatible with Kubernetes versions 1.27 to 1.30
+
+### New Features
```
pchico83, line 14: We should also document:
- Quotas at the namespace level
- Metrics tab for each resource in a namespace
- Support for [priorityClassName](https://www.okteto.com/docs/1.25/reference/okteto-manifest/#priorityclassname-string-optional) and [accessMode](https://www.okteto.com/docs/1.25/reference/okteto-manifest/#persistentvolume-object-optional) for volumes created by `okteto up`

src/content/release-notes.mdx
ifbyol: Yes, this was released on previous version, so we could skip it

src/content/release-notes.mdx
```diff
@@ -5,6 +5,43 @@ sidebar_label: Release notes
...
+- Support to enable/disable resource quotas for individual Namespaces  <!-- 8604 -->
+- Allow override of installer security context  <!-- https://github.com/okteto/app/pull/8608 -->
+
+### Improvements
+
+- Show more actionable feedback upon "failed to deploy okteto pipeline"…  <!-- 8521 -->
+- You can now list your Github App installations under Settings → Integrations  <!-- 8463 -->
+- Implemented a number of UI updates to ensure your Okteto experience stays smooth <!-- 8600, 8623, 8534, 8626 -->
+
+### Bug Fixes
```
ifbyol, line 31: There is a relevant fix in the CLI that it worths to be mentioned. It is a fix for Smart Builds. It would be something like:

`Fix in Smart Builds logic to properly calculate when to build a new image if the build context points to a parent folder`

src/content/release-notes.mdx
codyjlandstrom: @viddo Do you think it's worth mentioning?

pchico83 (approved):

src/content/release-notes.mdx
viddo: Yes, but not the way they are currently worded. The texts appears to be verbatim from the PR titles. I'd suggest to reword these based on the PR descriptions, something on the lines of:

```suggestion
- Fixed issue where interactive but hidden elements, such as links in closed deploy log stage, were wrongly accessible via keyboard navigation and screen readers <!-- 8522 -->
- Fixed issue preventing password managers from filling in the login token <!-- 8510 -->
```

Similarly, I think there's more UI bugfixes and improvements we done since time that are worthwhile to include in the release notes. I'll review internal changelog and add some more suggestions here & there probably.

---

BTW, as an aside, I think we could improve how we articulate PR titles/desc and auto-generate the release notes to avoid having to go back on so many items in this kind of docs PR. Let's talk about it in a retrospective and/or after this release to improve for next time?

src/content/release-notes.mdx
```diff
@@ -5,6 +5,44 @@ sidebar_label: Release notes
...
+This versions is compatible with Kubernetes versions 1.27 to 1.30
+
+### New Features
+
+- [Announcing Okteto CLI 3.0](https://www.okteto.com/blog/cli-three-release/) - upgrading to chart release 1.25 requires a CLI upgrade to 3.0
+- Announcing [Cloud Credentials](admin/cloud-credentials/index.mdx), a central location to manage your cloud provider credentials for `deploy`, `destroy`, and `test` remote operations  <!-- 8550 -->
+- Adding support for [resource quotas at the individual Namespace level](core/namespaces.mdx#configure-namespace-quotas). This allows administrators to set the maximum resources that can be used per Namespace
+- Enabling [Okteto Insights Dashboards](core/okteto-insights-dashboards.mdx) for all SaaS and BYOC users of Okteto <!-- 8627 -->
+- Added CLI, Chart, and Okteto App version to our API headers  <!-- 8545 -->
+- Added support to [enable/disable resource quotas for individual Namespaces](self-hosted/helm-configuration.mdx#quotas). The new default values do not enforce a resource limit  <!-- 8604 -->
```
ifbyol, line 21: @pchico83 Is this the same than `Adding support for [resource quotas at the individual Namespace level](core/namespaces.mdx#configure-namespace-quotas). This allows administrators to set the maximum resources that can be used per Namespace`? or 2 different things? I mean, we already had quotas for individual namespaces and we only allowed to disable them for some resources, right?

src/content/release-notes.mdx
```diff
@@ -5,6 +5,44 @@ sidebar_label: Release notes
...
+- Added support to allow override of [installer security context](self-hosted/helm-configuration.mdx#installer)  <!-- https://github.com/okteto/app/pull/8608 -->
+
+### Improvements
+
+- Support for [`priorityClassName`](reference/okteto-manifest.mdx#priorityclassname-string-optional) and [`accessMode`](reference/okteto-manifest.mdx#persistentvolume-object-optional) for volumes created by `okteto up`
+- Now showing more actionable feedback upon "failed to deploy okteto pipeline"…  <!-- 8521 -->
```
ifbyol, line 27: Are the three dots at the end of the sentence expected? Are they part of the `failed to deploy okteto pipeline" sentence?

viddo: No, could be reworded to be more clear too e.g.:
```suggestion
- Show more actionable feedback upon "failed to deploy okteto pipeline" error  <!-- 8521 -->
```

src/content/release-notes.mdx
pchico83: It's two different things, but I would unify them in a single feature:
- One feature is that we show the quotas on the Namespace UI. We already did it before, but they have been redesigned
- Another one is that, when you enable `quotas.resource.enabled`, each resource quota cana be enable/disabled independently

src/content/release-notes.mdx
```diff
@@ -5,6 +5,44 @@ sidebar_label: Release notes
...
+
+This versions is compatible with Kubernetes versions 1.27 to 1.30
+
+### New Features
+
+- [Announcing Okteto CLI 3.0](https://www.okteto.com/blog/cli-three-release/) - upgrading to chart release 1.25 requires a CLI upgrade to 3.0
+- Announcing [Cloud Credentials](admin/cloud-credentials/index.mdx), a central location to manage your cloud provider credentials for `deploy`, `destroy`, and `test` remote operations  <!-- 8550 -->
+- Adding support for [resource quotas at the individual Namespace level](core/namespaces.mdx#configure-namespace-quotas). This allows administrators to set the maximum resources that can be used per Namespace
+- Enabling [Okteto Insights Dashboards](core/okteto-insights-dashboards.mdx) for all SaaS and BYOC users of Okteto <!-- 8627 -->
+- Added CLI, Chart, and Okteto App version to our API headers  <!-- 8545 -->
```
viddo, line 20: What does "our" refer to more specifically? How does this affect/help customers? Asking because looking at referenced PR and parent issue, it appears to be more for internal usage than "new feature" that any customer would use, isn't it?

codyjlandstrom: Oddly enough we've received a few customer requests to have this version data so that our customer's platform team can have more information when debugging developer issues.
Good call out though on "our" I'll figure out a way to reword this to be more clear and show the value prop

src/content/release-notes.mdx
```diff
@@ -5,6 +5,42 @@ sidebar_label: Release notes
 id: release-notes
 ---
 
+## 1.25.0
+
+7 October 2024
+
+This versions is compatible with Kubernetes versions 1.27 to 1.30
+
+### New Features
```
pchico83, line 14: @codyjlandstrom should we announce Openshift support too?

codyjlandstrom: Good call out, updated!

PR #874, opened 2024-10-04
Move the "Deploy first app" tutorial our of the platform engineers ge…

pchico83:
…tting started guides

Playing with the "Get started" layout to give more visibility to the "Deploy your app" tutorial.


codyjlandstrom (approved):
New sidebar structure is much easier to follow. This is great

PR #875, opened 2024-10-07
Create 1.25 Version Release

codyjlandstrom:


pchico83:
@codyjlandstrom I think you are missing this step:
https://github.com/okteto/docs?tab=readme-ov-file#update-variables

codyjlandstrom:
> @codyjlandstrom I think you are missing this step: https://github.com/okteto/docs?tab=readme-ov-file#update-variables

Thanks for the catch. Fixed

pchico83:
> > @codyjlandstrom I think you are missing this step: https://github.com/okteto/docs?tab=readme-ov-file#update-variables
> 
> Thanks for the catch. Fixed

You need to update the variables file in the 1.25 `versioned` docs too

docusaurus.config.js
```diff
@@ -239,21 +239,26 @@ module.exports = {
...
           versions: {
             current: {
               // aka unreleased version in development
               // Remember to also update "unreleased" redirect if changing the value!
-              label: '1.25',
-              path: '1.25',
+              label: '1.26',
+              path: '1.26',
             },
             1.22: {
```
ifbyol, line 250: Is this correct? Shouldn't it be `1.25`?

codyjlandstrom: This is always one version ahead

ifbyol: I refer to the key. It says `1.22` which makes me think it is not right. So, in the `versions`  object, it seems there are elements with the key `1.22`

codyjlandstrom: I see what you're saying. It's odd that this compiles just fine. When I change the key to 1.25 it has a failure.
Figuring out what's going on...

PR #883, opened 2024-10-17
updating aws lambda guide

RinkiyaKeDad:


rberrelleza (approved):

src/content/tutorials/aws-lambda.mdx
```diff
@@ -7,66 +7,34 @@ id: aws-lambda
...
 
 - AWSCloudFormationFullAccess
 - IAMFullAccess
 - AWSLambda_FullAccess
 - AmazonAPIGatewayAdministrator
 - AmazonS3FullAccess
 
-Once the user has been created, [generate a set of access keys](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html), and save them file locally.
-
-In order for the installer to be able to use this keys, we need to add them as [Okteto Variables](core/okteto-variables.mdx). Use `AWS_ACCESS_KEY_ID` as the name for the variable that contains the access ID, and `AWS_SECRET_ACCESS_KEY` as the name for the variable that contains the access key.
```
codyjlandstrom, line 31: Should we leave this section as a backup in case someone doesn't want to use Cloud Credentials? I can't see why they wouldn't but is it worth showing another option?

rberrelleza: In general, I find it confusing when docs give me multiple options. That being said, using credentials it's a lot easier than roles tbh, so I can see how some people will prefer that. Maybe we can add it as an info block @codyjlandstrom ?

codyjlandstrom: Maybe lets remove it for now and create this as a community post? That may be the best spot for this information, then we don't clog up the onboarding guides

codyjlandstrom (approved):

PR #885, opened 2024-10-18
AirGap Networks installation guide

pchico83:
This is still work in progress, but I want to share it to visualize the expected results once all the work is done

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -246,6 +246,7 @@ module.exports = {
             'self-hosted/manage/buildkit-high-performance',
             'self-hosted/manage/backup',
             'self-hosted/manage/custom-resource-definitions',
+            'self-hosted/manage/airgap',
             'self-hosted/manage/troubleshooting',
             'self-hosted/manage/diagnostics',
             'self-hosted/manage/uninstall-okteto',
--- a/src/content/self-hosted/manage/airgap.mdx
+++ b/src/content/self-hosted/manage/airgap.mdx
@@ -1,0 +1,98 @@
+---
+title: AirGap Networks
+description: Installing Okteto in AirGap Networks
+sidebar_label: AirGap Networks
+id: airgap
+---
+
+import variables from '../../variables.json';
+import CodeBlock from '@theme/CodeBlock';
+import TiersList from "@theme/TiersList";
+
+# AirGap Networks<TiersList tiers="Enterprise Self-Hosted" />
+
+Okteto works on AirGap Networks, but it requires custom configuration.
+This guide has been crafted to help you overcome these hurdles, but we encourage you to [talk to us](https://www.okteto.com/get-demo/) to guide you during the installation.
+
+
+## Step 1: Disable Okteto Telemetry
+
+By default, Okteto sends [telemetry data](self-hosted/helm-configuration.mdx#telemetry) to help us understand usage patterns and prioritize improvements.
+In air-gapped environments, you can disable telemetry by adding the following to your Okteto Helm configuration:
+
+```yaml
+telemetry:
+  enabled: false
+```
+
+Disabling telemetry will automatically stop Okteto CLI analytics for all developers.
+
+## Step 2: Host Images in your Private Registry
+
+By default, all the images needed to deploy the Okteto Platform, and the images used by the Okteto CLI, are hosted in DockerHub.
+In air-gapped environments, you’ll need to host these images in your private registry.
+
+:::info
+This guide assumes that your cluster nodes have pull access to your private registry. If that's not the case, please follow your infrastructure provider's instructions on how to do it
+:::
+
+Push the following images to your private registry for the Okteto Chart installation:
+
+- **okteto/backend:{variables.chartVersion}**
+- **okteto/frontend:{variables.chartVersion}**
+- **okteto/buildkit:{variables.chartVersion}** and **okteto/buildkit:{variables.chartVersion}-rootless**
+- **okteto/registry:{variables.chartVersion}**
+- **okteto/pipeline-runner:{variables.chartVersion}**
+- **okteto/daemon:{variables.chartVersion}**
+- **registry.k8s.io/ingress-nginx/controller-chroot:{variables.nginxVersion}**
+- **ghcr.io/stakater/reloader:{variables.reloaderVersion}**
+- **bitnami/redis:{variables.redisVersion}**
+
+For each Okteto CLI version, push the following images to your private registry:
+
+- **okteto/okteto:{variables.cliVersion}**
+
+:::info
+For Okteto Chart {variables.chartVersion}, we recommend enforcing the usage of Okteto CLI {variables.cliVersion} in your developer's machines to reduce maintenance overhead
+:::
+
+## Step 3: Update Helm Configuration
+
+Once your images are pushed to your private registry, update your Okteto Helm configuration as follows:
+
+```yaml
+globals:
+  registry: <<your-registry-url>>
+
+cli:
+  image:
+    registry: <<your-registry-url>>
+
+ingress-nginx:
+  global:
+    image:
+      registry: <<your-registry-url>>
+
+okteto-nginx:
+  global:
+    image: <<your-registry-url>>
+
+reloader:
+  reloader:
+    deployment:
+      image:
+        name: <<your-registry-url>>/stakater/reloader
+
+redis:
+  global:
+    imageRegistry: <<your-registry-url>>
+```
+
+
+## Step 4: Install or Upgrade Okteto 
+
+Execute the following command to install or upgrade your Okteto instance:
+
+<CodeBlock language="bash">
+  {`helm upgrade --install okteto okteto/okteto -f config.yaml --namespace=okteto --create-namespace --version=${variables.chartVersion}`}
+</CodeBlock>
--- a/src/content/variables.json
+++ b/src/content/variables.json
@@ -2,5 +2,8 @@
   "kubernetesMinVersion": "1.27",
   "kubernetesMaxVersion": "1.30",
   "cliVersion": "3.1.0",
-  "chartVersion": "1.26.0"
+  "chartVersion": "1.26.0",
+  "nginxVersion": "v1.11.2",
+  "reloaderVersion": "v1.0.119",
+  "redisVersion": "7.4.0-debian-12-r1"
 }


jvc5546 (commented):
LGTM!

sidebars.js
```diff
@@ -245,6 +245,7 @@ module.exports = {
             'self-hosted/manage/buildkit-high-performance',
             'self-hosted/manage/backup',
             'self-hosted/manage/custom-resource-definitions',
+            'self-hosted/manage/airgap',
```
codyjlandstrom, line 248: Can we move this under the "Complete the installation section"?

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,115 @@
...
+id: airgap
+---
+
+import variables from '../../variables.json';
+import CodeBlock from '@theme/CodeBlock';
+import TiersList from "@theme/TiersList";
+
+# AirGap Networks<TiersList tiers="Enterprise Self-Hosted" />
+
+Okteto works on AirGap Networks, but it requires custom configuration.
```
codyjlandstrom, line 14:
```suggestion
Okteto works in air gapped network environments, but will require some custom configuration.
```

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,115 @@
...
+This guide has been crafted to help you overcome these hurdles, but we encourage you to [talk to us](https://www.okteto.com/get-demo/) to guide you during the installation.
```
codyjlandstrom, line 15:
```suggestion
This guide has been crafted to help you overcome these hurdles, but we encourage you to [talk to us](https://support.okteto.com/hc/en-us) to guide you during the installation.
```

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
+---
+title: AirGap Networks
```
codyjlandstrom, line 2:
```suggestion
title: Air-Gapped Networks
```

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+description: Installing Okteto in AirGap Networks
```
codyjlandstrom, line 3:
```suggestion
description: Installing Okteto in Air-Gapped Network Environments
```

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+sidebar_label: AirGap Networks
```
codyjlandstrom, line 4:
```suggestion
sidebar_label: Okteto in Air-Gapped Networks
```

src/content/self-hosted/manage/airgap.mdx
rberrelleza: I think we should remove this line entirely. Anyone doing an airgapped install knows that extra steps are required.  Could we instead talk about what we mean by air gap? (e.g. is a good example IMO https://docs.gitlab.com/ee/topics/offline/quick_start_guide.html)

src/content/self-hosted/manage/airgap.mdx
rberrelleza: We should also rephrase this, I don't think this kind of language inspires confidence. We shouldn't say that "it's hard", "it takes more effort", "there are hurdles".  I think we should simply mention that "for airgapped environments, we recommend that you reach out to our team to discuss your requirements and scenarios" or something like that. 

For the link, since the customer is not asking for a demo but for help/guidance, I think that https://www.okteto.com/contact/ is more appropriate.

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+import CodeBlock from '@theme/CodeBlock';
+import TiersList from "@theme/TiersList";
+
+# AirGap Networks<TiersList tiers="Enterprise Self-Hosted" />
+
+Okteto works on AirGap Networks, but it requires custom configuration.
+This guide has been crafted to help you overcome these hurdles, but we encourage you to [talk to us](https://www.okteto.com/get-demo/) to guide you during the installation.
+
+
+## Step 1: Disable Okteto Telemetry
```
rberrelleza, line 18: I think we should start with the more complex scenarios before going into the 'easy' stuff :P

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+```yaml
+telemetry:
+  enabled: false
+```
+
+Disabling telemetry will automatically stop Okteto CLI analytics for all developers.
+
+## Step 2: Host Images in your Private Registry
+
+By default, all the images needed to deploy the Okteto Platform, and the images used by the Okteto CLI, are hosted in DockerHub.
```
rberrelleza, line 32:
```suggestion
By default, all the images needed to deploy the Okteto Platform and the images used by the Okteto CLI are hosted in DockerHub.
```

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+In air-gapped environments, you’ll need to host these images in your private registry.
+
+### Push Images to your Private Registry
```
rberrelleza, line 35: I don't think this subtitle is needed since there's no other 'subtask' in step 2.

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+For simplicity, we recommend using the same repository names and tags in your private registry.
```
rberrelleza, line 37: I think it's better to remove this line so we don't give people ideas that there are other ways. It's better if we recommend a single way of doing things (in this case, to upload to the private registry with the same name/tag)

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+:::info
+Okteto doesn't support `imagePullSecrets`, so your cluster must be configured to have access to your Private Registry,
```
rberrelleza, line 40: I don't undersand what we mean by "Okteto doesn't support imagePullSecrets". Are we talking about the helm chart?

rberrelleza: If so, I think we should rephrase this positively as "This guide assumes that the nodes of your cluster have image pull access to your private registry. If that's not the case, please follow your infrastructure provider's instructions on how to do it"(or something like that)

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+- **okteto/frontend:{variables.chartVersion}**
+- **okteto/buildkit:{variables.chartVersion}**
+- **okteto/registry:{variables.chartVersion}**
+- **okteto/pipeline-installer:{variables.chartVersion}**
+- **okteto/pipeline-runner:{variables.chartVersion}**
+- **okteto/ingress-nginx:{variables.chartVersion}**
+- **okteto/reloader:{variables.chartVersion}**
+- **okteto/redis:{variables.chartVersion}**
+
+For each Okteto CLI version, push the following images to your private registry:
```
rberrelleza, line 57:
```suggestion
For each Okteto CLI version that you will enable for your users, you'll need to also push the following images to your private registry:
```

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+- **okteto/okteto:{variables.cliVersion}**
+- **okteto/dockerfile:{variables.cliVersion}**
+
+:::info
+For Okteto Chart {variables.chartVersion}, we recommend enforcing the usage of Okteto CLI {variables.cliVersion} in your developer's machines to reduce maintenance overhead
```
rberrelleza, line 63: can we include a link on how to do this? 

Also, we shouldn't recommend, we should tell them this is  a must, otherwise things won't work (e.g. if they let developers install liberally, the okteto/okteto:VERSION image won't be available on the cluster and it will fail. I think 'setting the supported CLI' versions should be a step on the guide. 

In general, with air gapped we need to give stronger direction than on public cloud because every small deviation/variation can break things and slow down the installation.

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+:::
+
+## Step 3: Update Helm Configuration
+
+Once your images are pushed to your private registry, update your Okteto Helm configuration as follows:
+
+<CodeBlock language="yaml">
+  {`globals:
```
rberrelleza, line 71: The more things the user needs to configure, the more error prone it is. Is there a way for globals to apply to all of these?

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+  registry: <<your-registry-url>>
+
+ingress-nginx:
+  image:
+    image: <<your-registry-url>>/okteto/ingress-nginx
+    tag: ${variables.chartVersion}
```
rberrelleza, line 77: to they need to specify chartVersion or just the image?

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+Execute the following command to install or upgrade your Okteto instance:
+
+<CodeBlock language="bash">
+  {`helm upgrade --install okteto okteto/okteto -f config.yaml --namespace=okteto --create-namespace --version=${variables.chartVersion}`}
+</CodeBlock>
+
+## Optional: Use Your Own Ingress Controller
+
+By default, Okteto deploy an [NGINX Ingress Controller](https://kubernetes.github.io/ingress-nginx/) to expose the Okteto Control Plane and the applications managed by Okteto.
```
rberrelleza, line 109:
```suggestion
By default, Okteto deploys [NGINX Ingress Controller](https://kubernetes.github.io/ingress-nginx/) to expose the Okteto Control Plane and the applications managed by Okteto.
```

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+Our NGINX Ingress Controller is usually exposed in a dedicated Load Balancer.
+
+Refer to [this community guide](https://community.okteto.com/t/how-do-i-configure-my-own-ingress-controller/887) for detailed instructions on disabling the installation of our Nginx Ingress Controller and exposing Okteto behind your own Ingress Controller.
```
rberrelleza, line 112:
```suggestion
Refer to [this community guide](https://community.okteto.com/t/how-do-i-configure-my-own-ingress-controller/887) for detailed instructions on disabling the installation of our Nginx Ingress Controller and exposing Okteto behind other Ingress Controllers.
```

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+Execute the following command to install or upgrade your Okteto instance:
+
+<CodeBlock language="bash">
+  {`helm upgrade --install okteto okteto/okteto -f config.yaml --namespace=okteto --create-namespace --version=${variables.chartVersion}`}
+</CodeBlock>
+
+## Optional: Use Your Own Ingress Controller
+
+By default, Okteto deploy an [NGINX Ingress Controller](https://kubernetes.github.io/ingress-nginx/) to expose the Okteto Control Plane and the applications managed by Okteto.
+Our NGINX Ingress Controller is usually exposed in a dedicated Load Balancer.
```
rberrelleza, line 110: Is this detail important for air-gapped installations?

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+Refer to [this community guide](https://community.okteto.com/t/how-do-i-configure-my-own-ingress-controller/887) for detailed instructions on disabling the installation of our Nginx Ingress Controller and exposing Okteto behind your own Ingress Controller.
+
+## Install Okteto without a Wildcard Certificate (optional)
```
rberrelleza, line 114: Is this truly optional on Air Gapped environments? I feel that this recommendation should be part of the Certificates page, since this might be the case even in other types of installations.

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+### Push Images to your Private Registry
+
+For simplicity, we recommend using the same repository names and tags in your private registry.
+
+:::info
+Okteto doesn't support `imagePullSecrets`, so your cluster must be configured to have access to your Private Registry, 
+:::
+
+**Required Images for Okteto Platform Installation**
```
rberrelleza, line 43:
```suggestion
**Required Docker images**
```

rberrelleza: I don't think we use "okteto platform" in the docs (e.g. see https://www.okteto.com/docs/get-started/install/), let's stay with the same terminology we already use in the docs.

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+Push the following images to your private registry for the Okteto Chart installation:
```
rberrelleza, line 45:
```suggestion
To install Okteto in an air-gapped environment, you'll need to push the following Docker images to your private registry before installing the Okteto helm chart.  

You must perform this action whenever you upgrade Okteto to a newer version.
```

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+- **okteto/backend:{variables.chartVersion}**
```
rberrelleza, line 47: I think we are missing the following images from here?
1. buildkit rootless
2. daemonset
3. helm (used by support bundle)

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,117 @@
...
+- **okteto/registry:{variables.chartVersion}**
+- **okteto/pipeline-installer:{variables.chartVersion}**
+- **okteto/pipeline-runner:{variables.chartVersion}**
+- **okteto/ingress-nginx:{variables.chartVersion}**
+- **okteto/reloader:{variables.chartVersion}**
+- **okteto/redis:{variables.chartVersion}**
+
+For each Okteto CLI version, push the following images to your private registry:
+
+- **okteto/okteto:{variables.cliVersion}**
```
rberrelleza, line 59: okteto/bin and busybox are missing from this I believe

rberrelleza (commented):
This is great! It's going to simplify things a lot for potential air gap customers

sidebars.js
rberrelleza: This is pretty advanced. Maybe we can start folding it into a different section? Maybe we should have an 'advanced scenarios' section or something.   

IMO Fine-tuning buildkit, doing backups, interacting with CRDs, and doing air-gapped installations are not 'completing your installation' but  a separate work stream. If we put everything, we risk that section of the docs becoming a hodge-podge of options without a straightforward narrative.  We could even consider separating installation tasks from app-level configuration from dev-level configuration in a way that maps the personas that typically perform those tasks?

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+By default, Okteto sends [telemetry data](self-hosted/helm-configuration.mdx#telemetry) to help us understand usage patterns and prioritize improvements
+In air-gapped environments, you can disable telemetry by adding the following to your Okteto Helm configuration:
+
+```yaml
+telemetry:
+  enabled: false
+```
+
+Disabling telemetry will automatically stop Okteto CLI analytics for all developers.
```
mnevadom, line 28: Im not sure but I think that we should disable it from the CLI too right? 
From my testings I recall it was enabled by default. Im still testing other things before this but I have it in my notes

src/content/self-hosted/manage/airgap.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+## Step 2: Host Images in your Private Registry
+
+By default, all the images needed to deploy the Okteto Platform and the images used by the Okteto CLI are hosted in DockerHub.
+In air-gapped environments, you’ll need to host these images in your private registry.
```
mnevadom, line 33: I was planning to add where to find all the images needed. Do you think is too much information for the guide? I mean, I wouldn't add every image we need, but I would add where to find them. 

For example, in the Chart Artifactory, in the values file, in the dependencies section for the nginx ...

mnevadom: I know we added all of them as a list later here, but that means that every time we add a new one we need to update this guide. For me its better if we teach them how to fish instead giving them fish :)

src/content/self-hosted/manage/airgap.mdx
mnevadom: I think it is yes. But Im not sure if we could give more details here. Its not just the option of enable/disable it, we could add the MetalLB option, or using ours but adding the machine IP to it: 

```
ingress-nginx:
  controller:
    service:
      externalIPs:
      - 10.132.0.2
```

But maybe thats something we want them to figure out internally...

mnevadom:
Another thing I would add is DNS problems. In the case of Elbit for example they had to talk to IT to be able to resolve Okteto domains. They needed to add all of them and I think they ended up adding the wildcard domain.

mnevadom:
Other things we should add is: 

- `syncthing`: the CLI downloads the binary and if they dont have access to it they need to add it manually and set an ENV variable with the version so that when the upgrade the CLI it doenst try to download a new one

- `docker/dockerfile:1`: Buildkit is going to try to download this image and it will fail if the cluster does not have access. They should add this one to the registry and add it to the Dockerfiles right? Im exactly at this point now with testing.

sidebars.js
pchico83: "Complete the installation section" is to have a production ready installation, this is more than that, it only applies to some scenarios. That's why sections like airgap, backups, CRDs, ... are under the "Operating" section

src/content/self-hosted/manage/airgap.mdx
pchico83: CLI is disabled automatically if you disable telemetry on the chart. Have you seen a different behavior?

src/content/self-hosted/manage/airgap.mdx
pchico83: Those are our own images, they don't change that much. We can list them for now and iterate if needed

src/content/self-hosted/manage/airgap.mdx
pchico83: We shouldn't mention specific controllers, this guide explains the general process:
https://community.okteto.com/t/how-do-i-configure-my-own-ingress-controller/887

src/content/self-hosted/manage/airgap.mdx
pchico83: It's more common on air-gapped. We can add it to the certificates section, but I call out here will also help

pchico83:
> * `syncthing`: the CLI downloads the binary and if they dont have access to it they need to add it manually and set an ENV variable with the version so that when the upgrade the CLI it doenst try to download a new one

@mnevadom do you have instructions on how to do this? This is not documented anywhere at the momemnt


> * `docker/dockerfile:1`: Buildkit is going to try to download this image and it will fail if the cluster does not have access. They should add this one to the registry and add it to the Dockerfiles right? Im exactly at this point now with testing.

It's already listed on the CLI images list

src/content/self-hosted/manage/airgap.mdx
mnevadom: I can't tell for sure cause I already disable it. But I think I installed Okteto with telemtry disabled and then I had to disable it in the CLI. 

I can test it if you want from scratch, but we could add something like "Make sure that you have analytics disabled in the CLI" ?

pchico83: Try to enable the CLI analytics `okteto analytics enabled` and it should send metrics if it's disabled at the helm level. It would be a critical bug otherwise

mnevadom:
@pchico83 would this work for synthing? https://community.okteto.com/t/how-do-i-force-the-okteto-cli-to-use-a-specific-version-of-syncthing/1284

In addition we could add that you need to download the binary with that version and put it in ~/.okteto

https://github.com/syncthing/syncthing

pchico83:
> @pchico83 would this work for synthing? https://community.okteto.com/t/how-do-i-force-the-okteto-cli-to-use-a-specific-version-of-syncthing/1284
> 
> In addition we could add that you need to download the binary with that version and put it in ~/.okteto
> 
> https://github.com/syncthing/syncthing

@mnevadom that works, thanks! I will add it to the doc

pchico83:
> @pchico83
> 
> Regarding docker/dockerfile:1, right now is hardcoded in the CLI. Even if we add it to the private registry won't work I think

@mnevadom right, it's part of the work we are doing for the next release

src/content/self-hosted/manage/air-gapped.mdx
```diff
@@ -0,0 +1,136 @@
...
+For the Okteto Chart {variables.chartVersion} installation, these are the images you need to push to your Private Registry:
+
+- **okteto/backend:{variables.chartVersion}**
+- **okteto/frontend:{variables.chartVersion}**
+- **okteto/buildkit:{variables.chartVersion}**
+- **okteto/buildkit:{variables.chartVersion}-rootless**
+- **okteto/registry:{variables.chartVersion}**
+- **okteto/pipeline-installer:{variables.chartVersion}**
+- **okteto/pipeline-runner:{variables.chartVersion}**
+- **okteto/daemon:{variables.chartVersion}**
```
mnevadom, line 47: I would add okteto/bin for the initContainer for okteto up. I added a FR to be able to add this image centralized and not in every manifest, I don't know if there is an ETA for it anyway :)

pchico83: okteto/bin and okteto/busybox will be unified with okteto/okteto

pchico83:
@codyjlandstrom as talked offline, I will merge the PR as it is to avoid merging hell, and you add your edits on a fresh PR

PR #892, opened 2024-10-29
Document how to expose BuildKit on a dedicated LB

pchico83:
Dument how to expose buildkit on a dedicated LB as a best practice

### DIFF
--- a/src/content/self-hosted/manage/buildkit-high-performance.mdx
+++ b/src/content/self-hosted/manage/buildkit-high-performance.mdx
@@ -7,15 +7,15 @@ sidebar_label: BuildKit High Performance
 
 # Configuring BuildKit for High Performance
 
-Configuring BuildKit for high-performance is essential for a smooth development experience with Okteto.
-Enhancing your BuildKit performance accelerates image builds, [test executions](testing/index.mdx) and deploy commands when using [Remote Execution](core/remote-execution.mdx).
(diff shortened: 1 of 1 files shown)


src/content/self-hosted/manage/buildkit-high-performance.mdx
```diff
@@ -68,6 +68,47 @@ buildkit:
...
+- More control over how build requests are distributed across BuildKit pods.
+- Offloading BuildKit traffic from the Nginx Ingress Controller reduces its network overhead, leading to improved performance for other services relying on the ingress controller
+
+**Cons**
+- There are extra costs associated with provisioning and maintaining an additional load balancer
+
+To expose BuildKit via a dedicated load balancer, add the following to your Okteto Helm configuration:
+
+```yaml
+buildkit:
```
ifbyol, line 86: I think you also need to specify, If not, I think we have a validation and fail with `"The configuration '.Values.buildkit.ingress.enabled' cannot be set to 'true' if '.Values.buildkit.service.type' is configured as 'LoadBalancer'."`

```
buildkit:
  ingress:
    enabled: false
```

pchico83: updated. we still need to test it in EKS

codyjlandstrom (approved):

PR #894, opened 2024-10-30
Add `defaultBackend.nameOverride` property

andreafalzetti:


andreafalzetti:
@pchico83 if there is a better place to document the migration, please let me know and I will update the PR

src/content/self-hosted/helm-configuration.mdx
```diff
@@ -323,6 +324,14 @@ tolerations:
   devPool: dev
 ```
 
+<a class="anchor anchorWithStickyNavbar_node_modules-@docusaurus-theme-classic-lib-theme-Heading-styles-module" id="manual-migration-steps-when-renaming-the-defaultbackend-service"></a>**Manual migration steps when renaming the defaultBackend service**
```
andreafalzetti, line 327: a bit hacky, but otherwise this heading appears in the table of contents int he right sidebar, which I think it's unnecessary. Other ideas, are very welcome

andreafalzetti: it works when used, but docurauris won't build it because it fails with 

```
Broken anchor on source page path = /docs/1.26/self-hosted/helm-configuration/:
11:14:18 AM:      -> linking to /docs/1.26/self-hosted/helm-configuration/#manual-migration-steps-when-renaming-the-defaultbackend-service
```

We need to find a different place for the steps and do a regular link

codyjlandstrom: Hmm I'm not sure where else makes sense to put them. I'm ok with this being a headline item since it could also be something that users search for.

andreafalzetti: @codyjlandstrom as discussed with @pchico83 this is an edge case that can only occurr for now, so we will not document the steps

codyjlandstrom (approved):

PR #898, opened 2024-11-04
Add docs on Garbage Collector Admin View

codyjlandstrom:


src/content/admin/cleanup.mdx
```diff
@@ -10,9 +10,38 @@ import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
 import TiersList from "@theme/TiersList";
 
-The Okteto Garbage Collector helps keep your infra costs under control by sleeping or deleting inactive environments.
+The Okteto Garbage Collector (GC) allows Admins to optimize infrastructure usage by automatically scaling down or deleting inactive Developer and Preview environments, helping control costs and improve resource efficiency.
 
-## Sleeping Resources
+## Enabling the Garbage Collector
+
+To enable the GC, navigate to **Admin → Garbage Collector** and click `Install` if it’s not already installed.
```
viddo, line 17: This is not correct. "Install" CTA just defers to these same docs here. Or should we link the install step to some other docs? (would require changing app code though)

@teresaromero @pchico83 How should we explain required actions an admin needs to take? How does one _uninstall/install_ GC exactly? I think this is related to the [Helm chart configuration](https://www.okteto.com/docs/1.26/self-hosted/helm-configuration/#gc) that I also think need a review and more consideration for the new feature, right? Also see/review [current changes](https://github.com/okteto/docs/pull/898/files#diff-5246fa871de346de64f5cbaf3735699dde71a326365ff143ce7219a80f38f839) below.

codyjlandstrom: Ok, thanks. I was basing this off the screenshots that showed the "Install" button without realizing that it just linked to the docs. I can edit the docs to reflect this, but we should also update the helm configuration to describe "installation"

src/content/admin/dashboard.mdx
```diff
@@ -207,4 +209,17 @@ If you choose to pull images from a private registry in addition to [Okteto's Co
 
 Dynamically configure Okteto's connection to AWS and GCP, eliminating the need for developers to manage static credentials.
 
-[Review this page for a full description of how to configure and start using Cloud Credentials →](admin/cloud-credentials/index.mdx)
\ No newline at end of file
+[Review this page for a full description of how to configure and start using Cloud Credentials →](admin/cloud-credentials/index.mdx)
+
+## Garbage Collector
+
+Configure sleep and delete periods for Okteto Developer Environments and Preview Environments from the UI.
```
pchico83, line 216: Same as before, today, "Developer Environments" is not used across product/docs

pchico83 (approved):

src/content/admin/dashboard.mdx
codyjlandstrom: Thanks, will update both of these

PR #903, opened 2024-11-06
update helm config adding release name rules and settings

andreafalzetti:
Adding documentation for the release name max length.

I've added `nameOverride` and `fullnameOverride` under the "Advanced Configuration" to reduce noise, as these are mostly used in edge-case scenarios where the release name is very long.

I've added the recently introduced validation rules for the release name max length.


codyjlandstrom (approved):
Sounds great

pchico83 (approved):

PR #909, opened 2024-11-08
[Resource Manager] Improve docs and add examples

andreafalzetti:
This PR addresses the following:

1. Rephrase 'Rolling Back Changes` to 'Exclude Resources'
2. Adding samples

### DIFF
--- a/src/content/admin/resource-manager.mdx
+++ b/src/content/admin/resource-manager.mdx
@@ -75,9 +75,38 @@ This tab is available for key resources, including Deployments, StatefulSets, Jo
 * To learn more about what metrics are available, see our [Namespace Documentation →](core/namespaces.mdx#resource-metrics-view)
 
 ## Exclude Resources
+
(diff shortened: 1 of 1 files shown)


src/content/admin/resource-manager.mdx
```diff
@@ -92,6 +120,21 @@ The Resource Manager relies on the Kubernetes Metric Server. If the Metric Serve
 
 ### Known Limitation: Support for ReplicaSet and CronJob
 
-The Resource Manager does not natively support optimizing ReplicaSets and CronJobs due to limitations in differentiating between them and their controllers (e.g., Deployments and Jobs). By default, the Resource Manager infers the original controller’s name from the pod’s name (e.g., extracting `${deploymentName}` from a pod named `${deploymentName}-644d69dbc4-fv5g5`).
```
andreafalzetti, line 95: I think we documented too much internal concepts that are not useful o the user, I've simplified and added an example. Let me know if you think we should explain more the technical reasons, and I can find a way to put it back.

codyjlandstrom (approved):

PR #915, opened 2024-11-12
add 1.26.1 release notes

andreafalzetti:


src/content/variables.json
```diff
@@ -2,5 +2,5 @@
   "kubernetesMinVersion": "1.27",
   "kubernetesMaxVersion": "1.30",
   "cliVersion": "3.1.0",
-  "chartVersion": "1.26.0"
+  "chartVersion": "1.26.1"
```
andreafalzetti, line 5: not sure if this should be updated or not? @codyjlandstrom

andreafalzetti: I've reverted the change, as in other PRs we don't update it either.

pchico83 (approved):

PR #916, opened 2024-11-12
Added documentation about how to configure IAM roles credentials for private registries using CRDs approach

ifbyol:
Added documentation to explain how to configure private registry credentials with IAM roles using CRDs. The current section was a bit mixed between static credentials and AWS ones, so I opted by creating sub sections with each type of credentials to make it clearer.

I added the changes up to `1.22` version which is the one where we included AWS IAM role configurations: https://www.okteto.com/docs/release-notes/#1220


codyjlandstrom:
Thanks for adding, @ifbyol 
I updated the copy across the entire CRDs doc. If this looks ok to you, I'll apply it to the older versions as well

codyjlandstrom:
@ifbyol Realizing this will have some overlap with https://github.com/okteto/docs/pull/919/
I think we can merge this one first and I'll resolve the conflicts with 919 afterwards

ifbyol:
> Thanks for adding, @ifbyol I updated the copy across the entire CRDs doc. If this looks ok to you, I'll apply it to the older versions as well

@codyjlandstrom Yeah. looks good!

codyjlandstrom (approved):

PR #919, opened 2024-11-13
Docs for Catalog and Cloud Credentials ReadOnly

teresaromero:
This PR adds the CRD example for Cloud Credentials and the link for the CRD guides in case the user wants to manage via CRD Catalog and Cloud Credentials


src/content/admin/cloud-credentials/index.mdx
```diff
@@ -23,3 +23,9 @@ Follow our guides below to learn how to configure your Cloud Credentials for sup
 - [Amazon Web Services](admin/cloud-credentials/aws.mdx)
 - [Google Cloud](admin/cloud-credentials/gcp.mdx)
 
+## Manage Cloud Credentials using CRDs
```
teresaromero, line 26: @codyjlandstrom i was doubting if adding this here or in each AWS, GCP credential description 🤔

src/content/self-hosted/manage/crds.mdx
```diff
@@ -99,3 +99,49 @@ spec:
 ```
 
 To apply the changes in a manual sceario, you can simply run the command `kubectl apply -f my-catalog-item.yaml` from your terminal to add the catalog item to the your Catalog in the cluster.
+
+## Cloud Credentials
```
teresaromero, line 103: Same here, should we document it separatly or wrap them under cloud creds?

src/content/admin/cloud-credentials/index.mdx
codyjlandstrom: I went ahead and added it. I think it makes sense to let people know that it's an option

src/content/self-hosted/manage/crds.mdx
codyjlandstrom: I think centralized here is great

src/content/admin/cloud-credentials/aws.mdx
```diff
@@ -182,6 +182,9 @@ This grants the IAM Role the necessary permissions to interact with S3.
 
 Now that you've created the IAM role, the final step is to configure the AWS credentials in Okteto.
 
+Cloud Credentials are mainly managed via the [Admin dashboard](admin/dashboard.mdx#cloud-credentials), but can also be managed via one of the CRDs installed by Okteto.
+Follow the steps below to add them via the Admin dashboard, or [follow this guide to add them as a CRD](self-hosted/manage/crds.mdx#cloud-credentials). Please note that if you add credentials using CRDs they will be displayed in the UI, but they can't be modified from the UI.
+
```
codyjlandstrom, line 187: @pchico83 Do you think it makes sense to add this CRDs call out to the cloud provider guides like this?

pchico83: For consistency, I would follow the same approach as [registry credentials](https://www.okteto.com/docs/1.27/admin/registry-credentials/). If we add the callout here, I would add it to the Registry Credentials doc, too.
Also, more than a call out, I would say the user has 2 options:
- Create it from the UI for a more interactive experience
- Create it using CRDs and GitOps to have more control on changes and show the info "read-only" in the UI
I would even add the CRDs explanation here, and keep the CRDs doc just as a list of our CRDs. Otherwise, that doc will grow a lot as we incorporate more things managed by CRDs

codyjlandstrom: Updated this to follow how we do it in registry credentials. I wanted to do this without replicating the CRD guide in multiple places. Think this works ok?

src/content/self-hosted/manage/crds.mdx
```diff
@@ -128,3 +132,55 @@ spec:
...
+Cloud Credentials are mainly managed via the [Admin dashboard](admin/dashboard.mdx#cloud-credentials), but can also be managed via one of the CRDs installed by Okteto: `dynamicconfigs.admin.okteto.com`.
+Using CRDs can be useful if you would like to add Cloud Credentials manually with a simple `kubectl` command or manage your Cloud Credentials using GitOps.
+
+:::note
+If you add credentials using CRDs they will be displayed in the UI, but they can't be modified from the UI. If you want to manage them from the UI, they must be created from there.
+:::
+
+For example, if you want to add a configuration for GCP Cloud Credential, you will need to create a resource like this:
+
+- `metadata.name`: `gcp-credentials-config`
```
ifbyol, line 148: I think there are 2 important things to remark related to Cloud credentials:
* In order to work, the name for each one has to be a specific one (`gcp-credentials-config` and `aws-credentials-config`)
* You can only specify one credentials per provider. It means 1 for GCP and 1 for AWS maximum. You can have one for each of them, but you cannot have more than 1 for each type

pchico83 (approved):

src/content/self-hosted/manage/crds.mdx
codyjlandstrom: Thanks. Added a note to clarify this

PR #922, opened 2024-11-14
Remove manual migration steps for default-backend

andreafalzetti:
We introduced a change that makes it easier to migrate the default-backend, and requires no manual changes.

I've tested migration from `1.26` to "latest" and backwards, and the manual steps were not required as expected, so I am proposing to remove these steps that add additional friction in an upgrade and are no longer needed.

I would keep them in the `1.26` docs, as the change will be available from `1.27` onwards, so migrating from `1.27` to `1.28` will no require manual steps.

### DIFF
--- a/src/content/release-notes.mdx
+++ b/src/content/release-notes.mdx
@@ -25,7 +25,7 @@ This version is compatible with Kubernetes versions 1.27 to 1.30
 ### Breaking Changes
 
 - **Hostname Length Limit**: Deploys now fail if a service hostname exceeds 63 characters, showing an error message. Previously, the dev environment deployed successfully even if the endpoint didn’t work. This may affect environments that deployed without issues before
-- **Helm Release Name Limit**: Helm release names are now limited in length. Use `defaultBackend.nameOverride` to shorten the default backend service name if needed. If changing this during an upgrade, [follow this guide as it may impact the installation](self-hosted/helm-configuration.mdx#manual-migration-steps-when-renaming-the-defaultbackend-service)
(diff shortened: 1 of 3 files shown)


pchico83 (approved):

src/content/release-notes.mdx
```diff
@@ -25,7 +25,7 @@ This version is compatible with Kubernetes versions 1.27 to 1.30
 ### Breaking Changes
 
 - **Hostname Length Limit**: Deploys now fail if a service hostname exceeds 63 characters, showing an error message. Previously, the dev environment deployed successfully even if the endpoint didn’t work. This may affect environments that deployed without issues before
-- **Helm Release Name Limit**: Helm release names are now limited in length. Use `defaultBackend.nameOverride` to shorten the default backend service name if needed. If changing this during an upgrade, [follow this guide as it may impact the installation](self-hosted/helm-configuration.mdx#manual-migration-steps-when-renaming-the-defaultbackend-service)
+- **Helm Release Name Limit**: Helm release names are now limited in length. Use `defaultBackend.nameOverride` to shorten the default backend service name if needed. If changing this during an upgrade, [follow this guide as it may impact the installation](/docs/1.26/self-hosted/helm-configuration.mdx#manual-migration-steps-when-renaming-the-defaultbackend-service)
```
andreafalzetti, line 28: this link does not exist yet, but it should exist once we release 1.27 and 1.26 becomes versioned.

andreafalzetti: @codyjlandstrom do you know a better way to link to that paragraph that will only exist in 1.26?

codyjlandstrom: I'm not quite sure I follow. We want to add the link to 1.26, but only after the release of 1.27?

PR #926, opened 2024-11-21
Add version tags to breaking change notes so they can be linked to

codyjlandstrom:


ifbyol (approved):

src/content/release-notes.mdx
```diff
@@ -22,7 +22,7 @@ This version is compatible with Kubernetes versions 1.27 to 1.30
 
 This version is compatible with Kubernetes versions 1.27 to 1.30
 
-### Breaking Changes
+### Breaking Changes - 1.26
```
pchico83, line 25: Instead of this, you can keep the title as it is, and customize the anchor link like we do [here](https://github.com/okteto/docs/blob/main/src/content/reference/okteto-manifest.mdx#deploy-remotely-recommended-deploy-remotely). I think keeping the title is cleaner

codyjlandstrom: Amazing. I didn't realize you could do that.

PR #942, opened 2024-12-09
Add initial 1.27 release notes

codyjlandstrom:


src/content/release-notes.mdx
```diff
@@ -5,6 +5,42 @@ sidebar_label: Release notes
...
+
+### New Features
+
+- We now allow the ability for admins of Okteto to [set the minimum supported CLI version for their team](admin/dashboard.mdx#command-line-cli)  <!-- 8829 -->
+- We've published the Okteto Manifest JSON Schema for inline suggestions and validation for creating and editing Okteto manifests within your code editor
+- Added support for overwriting CLI images at the Helm chart level
+- Shipped additional support for the installation of [Okteto in Air-Gapped Environments](self-hosted/manage/air-gapped.mdx) <!-- 8914 -->
+- Configured buildkit probes to only accept requests when buildkit is healthy  <!-- 8871 -->
+- Added support for configuring the Okteto control plane jobs TTL at the helm level  <!-- 8917 -->
+- feat: send cli image on graphql clustermetadata  <!-- 8914 -->
```
codyjlandstrom, line 26: Not sure I understand this. Does it need a mention?

src/content/release-notes.mdx
```diff
@@ -5,6 +5,42 @@ sidebar_label: Release notes
...
 
+## 1.27.0
+
+9 December 2024
+
+This version is compatible with Kubernetes versions 1.27 to 1.30
+
+### Breaking Changes
+
+- Remove code for our deprecated Quickstarts feature <!-- 8842 -->
```
pchico83, line 16: There is another breaking change about to overwrite the container images in the helm chart. I would link to the airgap guide saying that's the new way to overwrite images

src/content/release-notes.mdx
```diff
@@ -5,6 +5,42 @@ sidebar_label: Release notes
...
+
+### New Features
+
+- We now allow the ability for admins of Okteto to [set the minimum supported CLI version for their team](admin/dashboard.mdx#command-line-cli)  <!-- 8829 -->
+- We've published the Okteto Manifest JSON Schema for inline suggestions and validation for creating and editing Okteto manifests within your code editor
+- Added support for overwriting CLI images at the Helm chart level
```
pchico83, line 22: Can we link to the tech docs in our helm config reference?

src/content/release-notes.mdx
```diff
@@ -5,6 +5,42 @@ sidebar_label: Release notes
...
+- Shipped additional support for the installation of [Okteto in Air-Gapped Environments](self-hosted/manage/air-gapped.mdx) <!-- 8914 -->
+- Configured buildkit probes to only accept requests when buildkit is healthy  <!-- 8871 -->
+- Added support for configuring the Okteto control plane jobs TTL at the helm level  <!-- 8917 -->
```
pchico83, line 25: Can we link to the helm config docs explaining it?

src/content/release-notes.mdx
```diff
@@ -5,6 +5,42 @@ sidebar_label: Release notes
...
+- feat: send cli image on graphql clustermetadata  <!-- 8914 -->
+
+### Improvements
+
+- We've reorganized the [Admin Dashboard menu items](admin/dashboard.mdx) into groups for easier navigation  <!-- 8840 -->
+- Added readonly tags in the Dashboard to Catalog, Cloud Credentials, and Registry Credential items that were created via CRDs  <!-- 8886 -->
```
pchico83, line 31: I would say this is a feature. Also, "read-only" Registry credentials were implemented several releases ago, it's not a new feature

src/content/release-notes.mdx
```diff
@@ -5,6 +5,42 @@ sidebar_label: Release notes
...
+- Updated buildkit cacheRatio value to `0.8`  <!-- 8921 -->
```
pchico83, line 32: It's updated to 0.5

src/content/release-notes.mdx
```diff
@@ -5,6 +5,42 @@ sidebar_label: Release notes
...
+- feat: `wake`, `sleep` & `destroyall` rootless (PLAT-738)  <!-- 8919 -->
```
pchico83, line 33: In the next two items I would say that we have migrated all the images of the control plane to be rootless to improve the overall security of Okteto

src/content/release-notes.mdx
```diff
@@ -5,6 +5,42 @@ sidebar_label: Release notes
...
+This version is compatible with Kubernetes versions 1.27 to 1.30
+
+### Breaking Changes
+
+- Remove code for our deprecated Quickstarts feature <!-- 8842 -->
+
+### New Features
+
+- We now allow the ability for admins of Okteto to [set the minimum supported CLI version for their team](admin/dashboard.mdx#command-line-cli)  <!-- 8829 -->
+- We've published the Okteto Manifest JSON Schema for inline suggestions and validation for creating and editing Okteto manifests within your code editor
```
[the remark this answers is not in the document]
ifbyol: Agree, this would be very useful to have

src/content/release-notes.mdx
ifbyol: No, I don't think this should be mentioned. It is an internal detail of implementation for the images management improvements

src/content/release-notes.mdx
ifbyol: I would specify is the _default_ value

src/content/release-notes.mdx
ifbyol: Agree, I would use a single point for this. Because it affects to all the components

src/content/release-notes.mdx
```diff
@@ -5,6 +5,42 @@ sidebar_label: Release notes
...
+- Added support for overwriting CLI images at the Helm chart level
+- Shipped additional support for the installation of [Okteto in Air-Gapped Environments](self-hosted/manage/air-gapped.mdx) <!-- 8914 -->
+- Configured buildkit probes to only accept requests when buildkit is healthy  <!-- 8871 -->
+- Added support for configuring the Okteto control plane jobs TTL at the helm level  <!-- 8917 -->
+- feat: send cli image on graphql clustermetadata  <!-- 8914 -->
+
+### Improvements
```
ifbyol, line 28: We should mention also the `3.2` release for the CLI. relevant things to mention:
* CLI image (`okteto/okteto`) is now rootless
* You can define a variable `OKTETO_DEV_PERSISTENT_VOLUME_SIZE`, and it will change the default volume size for dev containers (it only applies if the manifest doesn't set specifically a size)
* Fix: Fixed a problem when deploying a compose with failed healthchecks. We were not taking into account the timeout operation, so the deploy operation was stuck forever

src/content/release-notes.mdx
```diff
@@ -5,6 +5,42 @@ sidebar_label: Release notes
...
+
+This version is compatible with Kubernetes versions 1.27 to 1.30
+
+### Breaking Changes
+
+- Remove code for our deprecated Quickstarts feature <!-- 8842 -->
+
+### New Features
+
+- We now allow the ability for admins of Okteto to [set the minimum supported CLI version for their team](admin/dashboard.mdx#command-line-cli)  <!-- 8829 -->
```
ifbyol, line 20: Should we call out here that this will start to apply from CLI `3.2` and above?

src/content/release-notes.mdx
ifbyol: @pchico83 Should we mention also that installer image is no longer needed? In case someone might be specifying that image for whatever reason

pchico83: yes, makes sense. we could also mention the cli images that have been combined in `okteto/okteto`

src/content/release-notes.mdx
ifbyol: I think including this breaking change is still pending

src/content/release-notes.mdx
ifbyol: After the latest changes and the discussion we had on Monday, we should remove the point about `okteto/okteto` image being rootless. We have reverted it

src/content/release-notes.mdx
```diff
@@ -5,6 +5,45 @@ sidebar_label: Release notes
...
+- Configured buildkit probes to only accept requests when buildkit is healthy  <!-- 8871 -->
+- Added support for configuring the Okteto control plane jobs TTL at the helm level  <!-- 8917 -->
+
+### Improvements
+
+- We've reorganized the [Admin Dashboard menu items](admin/dashboard.mdx) into groups for easier navigation  <!-- 8840 -->
+- For Catalog, Cloud Credentials, and Registry Credential items that were created via CRDs, we've added "read-only" tags in the Dashboard to avoid confusion on which items can be edited in the UI <!-- 8886 -->
+- Updated buildkit cacheRatio default value to `0.5`  <!-- 8921 -->
+- The Okteto installer image is no longer needed, binaries for the installer jobs are now installed from the CLI image <!-- 8767 -->
+- Okteto bin image and busybox
```
ifbyol, line 33: This seems incomplete. These images were merged with the CLI image to simplify the number of docker images to manage. I would prefix it with `[Okteto CLI 3.2.0]`

src/content/release-notes.mdx
codyjlandstrom: Do we have docs for this? It's not https://www.okteto.com/docs/1.27/self-hosted/helm-configuration/#jobs is it?

ifbyol: Yes, it is that one. That allows you to configure how long jobs created by Okteto control plane exists in the cluster after they finish the execution

pchico83 (approved):

PR #955, opened 2024-12-20
Added a clarification about the delete period of the GC settings

ifbyol:
Added a clarification in the delete period description to indicate the difference when the sleep period is enabled or not

### DIFF
--- a/src/content/admin/cleanup.mdx
+++ b/src/content/admin/cleanup.mdx
@@ -22,7 +22,7 @@ Once enabled, the GC automatically scales idle environments to zero and deletes 
 Admins can set custom sleep and delete periods for both Namespaces and Preview Environments:
 
 - **Sleep Period**: The amount of time, in hours, that an environment must remain idle before the GC scales it down to zero
-- **Delete Period**: The number of days an environment must stay in a sleeping state before the GC permanently deletes it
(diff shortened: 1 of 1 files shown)


src/content/admin/cleanup.mdx
```diff
@@ -22,7 +22,7 @@ Once enabled, the GC automatically scales idle environments to zero and deletes
 Admins can set custom sleep and delete periods for both Namespaces and Preview Environments:
 
 - **Sleep Period**: The amount of time, in hours, that an environment must remain idle before the GC scales it down to zero
-- **Delete Period**: The number of days an environment must stay in a sleeping state before the GC permanently deletes it
+- **Delete Period**: The amount of time (in hours or days) that an environment must reamin idle before the GC permanently deletes it. If the sleep period is set, this represents the amount of time that the environment must stay in sleeping status before delete it. If sleep period is disabled, this represents the amount of time since the last update in the environment.
```
codyjlandstrom, line 25:
```suggestion
- **Delete Period**: The amount of time (in hours or days) that an environment must reamin idle before the GC permanently deletes it. If the sleep period is set, this represents the amount of time that the environment must stay in sleeping status before it is deleted. If sleep period is disabled, this represents the amount of time since the last update in the environment before it is deleted.
```

codyjlandstrom (approved):

PR #967, opened 2025-01-10
Add skipIfNoFileChanges to docs

maroshii:


src/content/reference/okteto-manifest.mdx
```diff
@@ -1134,6 +1135,12 @@ hosts:
     ip: 127.0.0.1
 ```
 
+#### skipIfNoFileChanges (boolean, optional)
+
+If set to true, the test will be skipped if no files have changed since the last test run. This is useful to speed up the test execution when no changes have been made.
+Files added to the `.oktetoignore` file will not be considered for this check.
```
pchico83, line 1141: Can we add that the default value is `false`?

pchico83 (approved):

ifbyol (approved):

PR #968, opened 2025-01-13
Create Docs 1.28 Version Release

codyjlandstrom:


ifbyol (commented):
I think we would need to update variables.json file too with the latest versions

pchico83:
> I think we would need to update variables.json file too with the latest versions

@codyjlandstrom it needs to be updated in the version-1.28/variables.json file

ifbyol:
> > I think we would need to update variables.json file too with the latest versions
> 
> @codyjlandstrom it needs to be updated in the `versioned-1.28/variables.json` file

But it need to be updated in `variables.json` in non-versioned content, right? I don't know if there we specify the latest stable or the next one, but `1.27` of the chart should be not specified there, right?

pchico83 (approved):

PR #972, opened 2025-01-17
Add notes on browser compatibility

codyjlandstrom:
@rlamana Could you verify that these are the versions that we want to support?

- **Google Chrome**: Latest 2 major versions
- **Mozilla Firefox**: Latest 2 major versions
- **Microsoft Edge**: Latest 2 major versions
- **Safari**: Latest 2 major versions (macOS only)

### DIFF
--- a/src/content/get-started/install/index.mdx
+++ b/src/content/get-started/install/index.mdx
@@ -26,10 +26,10 @@ If you need help installing Okteto on your Kubernetes cluster, we have a playlis
 ## Supported Browsers
 
 The Okteto Dashboard is optimized for the following browsers:
-- **Google Chrome**: Latest 2 major versions
-- **Mozilla Firefox**: Latest 2 major versions
-- **Microsoft Edge**: Latest 2 major versions
-- **Safari**: Latest 2 major versions (macOS only)
+- **Google Chrome**: Last 8 versions
+- **Mozilla Firefox**: Last 8 versions
+- **Microsoft Edge**: Last 8 versions
+- **Safari**: Last 2 major versions (macOS only)
 
 For users installing Okteto in air-gapped environments, ensure that all users have one of the supported browser versions to avoid compatibility issues.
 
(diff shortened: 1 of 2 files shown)


rlamana:
We have configured "[browserlist](https://github.com/okteto/app/blob/3532d06b1f3393b345252145f6c9662fdd276358/frontend/package.json#L178-L179)" (which generated JS code with retro-compatibility) to support the last 8 versions for the main browsers, that doesn't necessarily mean that we support the last 2 major versions. We could change the configuration to make sure we do support exactly the 2 major versions, but I would need to check how this impacts our generated code, for example the bundle size. 

Let me check that change before adding this to the documentation.

src/content/get-started/install/index.mdx
```diff
@@ -23,6 +23,16 @@ If you need help installing Okteto on your Kubernetes cluster, we have a playlis
 
 <iframe width="560" height="315" src="https://www.youtube.com/embed/videoseries?list=PLVcXGaCKg-Y0XGujg_G7vCIGd_JEAQ4Sq" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe>
 
+## Supported Browsers
+
+The Okteto Dashboard is optimized for the following browsers:
+- **Google Chrome**: Latest 2 major versions
+- **Mozilla Firefox**: Latest 2 major versions
+- **Microsoft Edge**: Latest 2 major versions
+- **Safari**: Latest 2 major versions (macOS only)
```
rlamana, line 32: @codyjlandstrom Most browsers don't use the classic semantic versioning based on MAJOR.MINOR.PATCH. So that's the reason why we were supponting only the 8 last versions. In the case of Safari, they do use semantic versioning and it makes sense to use the "Last 2 major versions", it actually supports a couple more versions than using just the "last 8 versions" for this browser, so I will adjust it in browserslist.

```suggestion
The Okteto Dashboard is optimized for the following browsers:
- **Google Chrome**: Last 8 versions
- **Mozilla Firefox**: Last 8 versions
- **Microsoft Edge**: Last 8 versions
- **Safari**: Last 2 major versions (macOS only)
```

The bundle size doesn't change if we do that change to Safari only.

rlamana: This is our configuration and our browsers support according to "browserlist":
Screenshot 2025-01-22 at 11 32 06

rlamana: Browserslist changes in this PR: https://github.com/okteto/app/pull/9073

rlamana (approved):

PR #980, opened 2025-02-02
1.29 release notes

codyjlandstrom:


src/content/release-notes.mdx
```diff
@@ -7,26 +7,28 @@ id: release-notes
...
 
-### Breaking Changes {#breaking-changes-1.29}
-
-TBC
 
 ### New Features {#new-features-1.29}
 
-TBC
+- Added support for [Kubernetes 1.31](https://kubernetes.io/blog/2024/08/13/kubernetes-v1-31-release/)  <!-- 9081 -->
+- Personal Namespaces can now be [included in your Garbage Collection Policy](admin/cleanup.mdx#applying-garbage-collection-to-personal-namespaces) <!-- 9056 -->
```
ifbyol, line 19: Should we mention here the exceptions to set the expectations? talking about namespaces not being deleted, only emptied, and PVCs are not deleted.

In that way, users have the main summary in this item, instead of having to go to the details page to know those relevant things. I think there are relevant enough to be mentioned here. But it's my personal opinion

src/content/release-notes.mdx
```diff
@@ -7,26 +7,28 @@ id: release-notes
...
+- Okta users can now configure automatic deprovisioning of user accounts through Okteto  <!-- 9064 -->
```
ifbyol, line 20: Are we going to include a link to the documentation where we explain how to configure it?

src/content/release-notes.mdx
```diff
@@ -7,26 +7,28 @@ id: release-notes
...
 
 ### Improvements {#improvements-1.29}
 
-TBC
+- Improved the retry mechanism in `okteto up` to handle overnight disconnections more gracefully by treating syncthing timeouts as transient after a successful initial execution  <!-- 9069 -->
```
ifbyol, line 24: I wouldn't mention this for now. This was something we did specifically for one customer and it is behind an unsupported feature flag. We should discuss when to make GA and document it properly then

src/content/release-notes.mdx
```diff
@@ -7,26 +7,28 @@ id: release-notes
...
+- Improve warnings when a CLI user has a different version than within the accepted range set by the Okteto Admin
 
 ### Bug Fixes {#bug-fixes-1.29}
 
-TBC
+- Fixed an issue where [Divert](reference/okteto-manifest.mdx#divert) would not work with Docker Compose as intended
+- Restored & fixed an issue with Buildkit running rootless <!-- 9070 -->
+- Fixed an issue where Okteto failed to inject the ingress-nginx controller’s private IP when the service name was too long
```
ifbyol, line 31:
```suggestion
- Fixed an issue where Okteto failed to inject the ingress-nginx controller’s private IP in Okteto components when the service name was too long
```

src/content/release-notes.mdx
```diff
@@ -7,26 +7,28 @@ id: release-notes
...
 
-TBC
+- Improved the retry mechanism in `okteto up` to handle overnight disconnections more gracefully by treating syncthing timeouts as transient after a successful initial execution  <!-- 9069 -->
+- Improve warnings when a CLI user has a different version than within the accepted range set by the Okteto Admin
 
 ### Bug Fixes {#bug-fixes-1.29}
 
-TBC
+- Fixed an issue where [Divert](reference/okteto-manifest.mdx#divert) would not work with Docker Compose as intended
+- Restored & fixed an issue with Buildkit running rootless <!-- 9070 -->
```
ifbyol, line 30: @jpf-okteto Was this something external to customers or only internal? Does it worth to mention it in the release notes?

src/content/release-notes.mdx
```diff
@@ -7,26 +7,28 @@ id: release-notes
...
-TBC
 
 ### New Features {#new-features-1.29}
 
-TBC
+- Added support for [Kubernetes 1.31](https://kubernetes.io/blog/2024/08/13/kubernetes-v1-31-release/)  <!-- 9081 -->
+- Personal Namespaces can now be [included in your Garbage Collection Policy](admin/cleanup.mdx#applying-garbage-collection-to-personal-namespaces) <!-- 9056 -->
+- Okta users can now configure automatic deprovisioning of user accounts through Okteto  <!-- 9064 -->
 
 ### Improvements {#improvements-1.29}
```
ifbyol, line 22: As improvement, I would mention that the `okteto up` command now builds only the necessary images to be used in the process instead of build all the images. This is an improvement in the CLI

src/content/release-notes.mdx
ifbyol: But that was before (not since this version-1.29), and we didn't support Kubernetes `1.31` before this version, right? So, this was affecting customers, right?

src/content/release-notes.mdx
codyjlandstrom: I tried doing this, but I thought it was too long. Was thinking of recording a loom to send with the release notes to announce this and briefly cover. What do you think?

src/content/release-notes.mdx
codyjlandstrom: Debating waiting to include this at all until the UI is ready

src/content/release-notes.mdx
codyjlandstrom: Amazing! Added

src/content/release-notes.mdx
ifbyol: Yeah, we can document that if we find it relevant @codyjlandstrom . We were adding a deprecated annotation in BuildKit statefulset on kubernetes `1.30`

src/content/release-notes.mdx
ifbyol: Yeah, that could be also another good option

src/content/release-notes.mdx
codyjlandstrom: Sorry I got a bit lost by the conversation here. Should we update this line to say "We have added a deprecated annotation in BuildKit statefulset on kubernetes 1.30?"

pchico83 (approved):

ifbyol (approved):

PR #990, opened 2025-02-13
DEV-863: Remove support of Kubernetes version 1.28

ifbyol:
Changes to reflect that `1.30` version of Okteto will not support Kubernetes version `1.28`. 

@codyjlandstrom Do we want to mention something about now supporting one less version of Kubernetes per release, or it is fine just the regular removal notice?

### DIFF
--- a/src/content/release-notes.mdx
+++ b/src/content/release-notes.mdx
@@ -5,6 +5,33 @@ sidebar_label: Release notes
 id: release-notes
 ---
 
+## 1.30.0
+
+XX March 2025
+
+This version is compatible with Kubernetes versions 1.29 to 1.31 \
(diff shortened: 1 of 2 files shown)


codyjlandstrom:
> @codyjlandstrom Do we want to mention something about now supporting one less version of Kubernetes per release, or it is fine just the regular removal notice?

Yes we need to send a customer notification for this. Let me add it to the release note plan - we should also publish this policy somewhere in our docs

ifbyol:
> > @codyjlandstrom Do we want to mention something about now supporting one less version of Kubernetes per release, or it is fine just the regular removal notice?
> 
> Yes we need to send a customer notification for this. Let me add it to the release note plan - we should also publish this policy somewhere in our docs

@codyjlandstrom Ok, is that something that should block this PR?

PR #992, opened 2025-02-19
Fine tune configuration for ArgoCD ignore

ifbyol:
Adding a rule to ignore some fields in the `okteto-bot` service account created as part of the installation, and also improved the current rule for the mutatingwebhookconfiguration, as we had to manually add more lines if we include more webhooks on it. I also took the chance to create specific rules for validating and mutating webhooks (previous rule wasn't specifying the kind, so it was considering everything under the group `admissionregistration.k8s.io`

### DIFF
--- a/src/content/self-hosted/manage/argocd.mdx
+++ b/src/content/self-hosted/manage/argocd.mdx
@@ -80,21 +80,27 @@ ignoreDifferences:
       - '/data/tls.crt'
       - '/data/tls.key'
 
-  # Webhook cabundles patched by Okteto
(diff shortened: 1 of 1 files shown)


codyjlandstrom:
@ifbyol Cool if I add these to the current docs version as well?

ifbyol:
> @ifbyol Cool if I add these to the current docs version as well?

@codyjlandstrom It is already added to the current version, isn't it? It is in the default folder, but also in `version-1.29`. Or I'm not getting what you asking 😅

PR #996, opened 2025-03-01
Add tutorial on using Google Cloud Secret Manager with Okteto

codyjlandstrom:
Adds a tutorial showing how to inject secrets stored in Google Cloud Secret Manager into Okteto Development and Preview Environments, based on the [okteto-community/gcp-secret-manager](https://github.com/okteto-community/gcp-secret-manager) example.

Changes since the original draft:
- Moved from `admin/integrations` to the Tutorials section per review feedback
- Renamed to "Secret Manager" (singular) to match the GCP product name
- Rewritten to follow the tested approach from the community example (service account key stored as Admin Variables), with Workload Identity Federation linked as the keyless alternative instead of being a prerequisite


ifbyol (commented):
I would say this is a tutorial, more than an integration to be set in the admin documentation. Adding it here might give the impression that it can be used to manage okteto installation secrets or the CRDs one.

src/tutorials/gcp-secret-manager.mdx
```diff
@@ -0,0 +1,102 @@
+---
+title: Using Google Cloud Secret Manager with Okteto
```
rberrelleza, line 2: I think we should do this with https://www.okteto.com/docs/1.43/admin/cloud-credentials/ instead.

codyjlandstrom:
Closing. Ramiro's review points at doing this through Cloud Credentials (admin/cloud-credentials) instead of a standalone Secret Manager tutorial, which is a rewrite rather than a revision. If someone picks this up, start from the okteto-community/gcp-secret-manager example and the Cloud Credentials page rather than this branch.

PR #998, opened 2025-03-05
Add Okta User De-provisioning to the CRDs guide

codyjlandstrom:


src/content/self-hosted/manage/crds.mdx
```diff
@@ -188,4 +188,35 @@ To apply the changes manually, run the command `kubectl apply -f aws-cloud-crede
...
+- `metadata.name`: Unique identifier for the deprovisioning rule
+- `spec.oktaEventsAuthSecret`: Reference to the Okta Event Hook Token used in your Okta event hook configuration
+
+```yaml
+apiVersion: admin.okteto.com/v1
+kind: OktaUserDeprovisioning
+metadata:
+  name: okta-user-deprovisioning
+  namespace: okteto
+spec:
```
codyjlandstrom, line 209: @ifbyol Could I get your review here on the exact value for the resource?

ifbyol: Yeah, sure. On one side, you need to create a secret like the following one:

```
apiVersion: v1
data:
  eventsAuthSecret: ...
  eventsEnabled: ...
kind: Secret
metadata:
  name: <secret-name>
  namespace: okteto
type: Opaque
```

That secret should contain 2 keys with the information:
* `eventsEnabled`. It should be a boolean indicating that the configuration is enabled (or not). If the value is `false`, the feature is disabled
* `eventsAuthSecret`. It is the webhook authorization secret specified when configuring the event hook in Okta.

Once you have the secret, you have to create a CR like the following one:

```
apiVersion: admin.okteto.com/v1
kind: DynamicConfig
metadata:
  name: okta-config
  namespace: okteto
secretRef: <secret-name>
```

The field `secretRef` is a field at first level (there is no spec in this case), and it should contain the secret name created in the previous step

codyjlandstrom: @ifbyol Thanks for this! Could you review my updates and make sure that sounds ok?
Do we need to mention any encoding for the secret?

ifbyol (approved):
It looks good to me! Replied about the open thread, but I don't think it something blocker

src/content/self-hosted/manage/crds.mdx
codyjlandstrom: Great thanks. Added some clarifications in so it's clear. Mind double checking those really fast. Otherwise should be good

pchico83 (approved):

PR #1007, opened 2025-03-24
Fix formatting of appId in GitHub installation instructions

jLopezbarb:
Signed-off-by: Javier Lopez <javier@okteto.com>

Add double quotes in appId. It could be converted into a exponential and fail. Double quoting makes helm treat it like a string


ifbyol (approved):
Could you do it also from the current active version of the docs?

codyjlandstrom (approved):

PR #1048, opened 2025-06-24
Add new feature flags for automatic down and service synchronization …

jLopezbarb:
…in feature-flags.mdx


src/content/reference/feature-flags.mdx
```diff
@@ -28,6 +29,7 @@ Below is a list of those variables that you can use to leverage these newer and
...
 | OKTETO_LOCAL_REGISTRY_STORE_PRIORITY_ENABLED     | Gives more priority to the local registry credentials than the Okteto Registry Credentials                                                                                                                                      | `false`                |
 | OKTETO_PARALLEL_WAKE_UPS_FOR_DEVENVS             | Specifies if the dev environments within a namespace should wake up in parallel or not                                                                                                                                          | `false`                  |  
+| OKTETO_SERVICES_UP_WAIT                          | If set, waits for the services to be synchronised before running the commands                                                                                                                                                   | `false`                  |
```
ifbyol, line 32: Hmm, I think this might be confusing. It waits for the files to be synchronized to start the services defined under the `services` section, right?

By `waits for the services to be synchronised` i might not see that it refers to the code synchronization

codyjlandstrom:
```suggestion
| OKTETO_SERVICES_UP_WAIT                          | If set, waits for file synchronization to complete before running the commands                                                                                                                                                   | `false`                  |  
```

codyjlandstrom: Is this more accurate @ifbyol ?

ifbyol: @codyjlandstrom  I think it is important to highlight that it only affects to the services defined under the [`services` section](https://www.okteto.com/docs/reference/okteto-manifest/#services-object-optional). The main container was already waiting until the files were synched, but those services not. In those services is where there is the change of behavior

ifbyol: I think it sounds better. I like it!

PR #1062, opened 2025-08-04
Add "okteto/agent" to the list of images that must be hosted

pchico83:


pchico83:
> This should be merge for `1.36`, right? Not for the goingo `1.35`, right?

Yes, this is for 1.36

PR #1094, opened 2025-09-19
Upgrade Docusaurus and resolve webpack deprecation warning

codyjlandstrom:
- Upgrade all Docusaurus packages to latest 3.8.1
- Remove outdated docusaurus-plugin-copy causing webpack warnings
- Remove conflicting docusaurus v1.14.7 package
- Migrate static assets to modern static/ folder approach
- Move YAML config files from content to static directories

Fixes webpack DEP_WEBPACK_COMPILATION_ASSETS deprecation warning.

🤖 Generated with [Claude Code](https://claude.ai/code)


static/self-hosted/aks/config.yaml
```diff
@@ -0,0 +1,13 @@
+license: 
+
+subdomain:
```
pchico83, line 3: @codyjlandstrom why are we creating these new static files as part of the docusaurus upgrade?

PR #1104, opened 2025-10-01
Fix variables rendering in helm configuration installer section

codyjlandstrom:


codyjlandstrom:
We weren't importing the variables on those older pages either. I think this wasn't triggering any build errors because the variable reference was encapsulated in the code block. Something to watch out for in the future!

PR #1106, opened 2025-10-01
fix broken build

rberrelleza:
fixing broken link in tutorial that's blocking the build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,4 @@
-# syntax=docker/dockerfile:1.7-labs
-
-FROM node:20 as dev
+FROM node:22 as dev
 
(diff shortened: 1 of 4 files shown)


codyjlandstrom:
There are a number of issues here that I think would be solved by a rebase on the build for the failed PR. Could you try that first?

rberrelleza:
> There are a number of issues here that I think would be solved by a rebase on the build for the failed PR. Could you try that first?

branch is updated to latest, the builds failed due to a netflify configuration. I'll try to rerun the checks, but the deploy preview already runs in netlify

codyjlandstrom (approved):

PR #1155, opened 2026-02-04
docs: expand Divert setup, shared environments, and troubleshooting

codyjlandstrom:
Expands the Divert docs based on the feedback from the initial release, and reconciles the branch with everything that landed on main since February.

## What this adds

- **A setup path on the core concepts page** (`core/divert.mdx`): prerequisites by role, driver configuration, shared environment creation, manifest configuration, and verification. Each step links to the detailed guide instead of repeating it.
- **Setting Up Shared Environments**: how to create the shared namespace (Global Preview Environment vs regular Namespace), shared vs per-developer databases (with the Volume Snapshots tie-in), the hybrid pattern, and verification steps.
- **Troubleshooting, split by audience**: developer issues (header propagation, routing verification, a diagnostic checklist) live in Using Divert; admin issues (driver configuration, ingress controllers, sidecar injection) live in Configure Divert. The core page links to both instead of carrying a third copy.

## Corrections from the original draft

- **Linkerd is documented as optional for the nginx driver.** The original draft said it was required. Without Linkerd, the nginx driver routes at the ingress level; installing Linkerd extends header-based routing to service-to-service calls. This matches the CLI behavior and the chart.
- Removed a `divert.debug` manifest field that does not exist in the CLI schema.
- Removed an `api.logLevel` Helm value that does not exist in the chart.
- Dropped the inline Istio installation instructions (they duplicated the newer Istio installation guide and pointed at the old sample repo). The page links to the guide instead.
- Dropped the `versioned_docs/version-1.40` copies. 1.47 is current, so this content targets `src/content/` only.

## Merge with main

Merged main and kept everything that landed since this branch was opened: the Istio installation guide (#1177), the ingress controllers architecture section (#1187), the `okteto-community` sample repo URLs (#1178, #1196), and the `virtualServices`/`hosts` field documentation (#1197). Also fixed two pre-existing broken relative links in the Configure Divert "Next Steps" section.

Style pass done per `STYLE_GUIDE.md`. `yarn build` passes.


src/content/self-hosted/install/divert/linkerd-installation.mdx
ifbyol, line 89: There is some other configuration missed. Some changes in the okteto-nginx has to change, and a feature has to be enabled. Please, check the notion document, all the config changes needed are there

ifbyol, line 100: This is not correct. Adding this would imply that every namespace within Okteto would have the sidecard injection, but this is only needed in the shared environments, not in the rest of cases

src/content/self-hosted/install/divert/linkerd-installation.mdx
```diff
@@ -1,21 +1,21 @@
...
-This guide covers installing Linkerd to enhance Divert's routing capabilities when using the nginx driver. Linkerd provides service mesh functionality that enables more sophisticated traffic routing based on HTTP headers.
+This guide covers installing Linkerd to enable Divert's service-to-service routing when using the nginx driver. Linkerd provides the service mesh functionality that enables traffic routing based on HTTP headers between services.
 
 :::warning For nginx Driver Only
 Linkerd is **only** for use with the **nginx driver**. If you're using the istio driver, do not install Linkerd - Istio already provides service mesh capabilities. See [Configure Divert](index.mdx) for the distinction between drivers.
 :::
 
 :::info
-Linkerd is **optional** for the nginx driver. The basic Divert functionality works without it, but Linkerd enables enhanced service-to-service routing within the mesh.
+Linkerd is **optional** for the nginx driver. Without Linkerd, Divert routes traffic at the ingress level. Installing Linkerd extends header-based routing to service-to-service calls inside the cluster.
```
ifbyol, line 18: I'm not sure we should promote the usage of the nginx driver without Linkerd. it is easy that at some point some scenario where service-to-service communication is needed, and stops working, and people might think that Divert is not working fine

src/content/self-hosted/install/divert/linkerd-installation.mdx
ifbyol, line 109: Maybe it is important to highlight that this is only needed for shared environments. The rest of namespace doesn't require to do this

src/content/self-hosted/install/divert/linkerd-installation.mdx
```diff
@@ -126,7 +126,7 @@ Once configured by administrators, Linkerd works **transparently for all develop
 
 1. **Header-based routing**: Linkerd routes requests based on the `baggage` header at the service mesh level
 2. **Automatic retries**: Failed requests are automatically retried to the correct service
-3. **Load balancing**: Intelligent load balancing across service instances
+3. **Load balancing**: Latency-aware load balancing across service instances
 4. **mTLS**: Automatic mutual TLS between services for enhanced security
 5. **Observability**: Detailed metrics and tracing for diverted traffic (visible to administrators)
```
ifbyol, line 131: I wouldn't mention this for now, we didn't do any check on what is visible or not, and how they could be integrated with Okteto.

My fear is that if we mention this, people might expect having those metrics accessible somewhere, and right now is not possible

src/content/self-hosted/install/divert/linkerd-installation.mdx
```diff
@@ -126,7 +126,7 @@ Once configured by administrators, Linkerd works **transparently for all develop
 
 1. **Header-based routing**: Linkerd routes requests based on the `baggage` header at the service mesh level
 2. **Automatic retries**: Failed requests are automatically retried to the correct service
-3. **Load balancing**: Intelligent load balancing across service instances
+3. **Load balancing**: Latency-aware load balancing across service instances
```
ifbyol, line 129: From where are we getting this? Divert doesn't do anything for automatic retries nor Load Balancing.

I would only mention that Linkerd is the service mesh that allows do the header-based routing in a service-to-service communication.

src/content/self-hosted/install/divert/linkerd-installation.mdx
ifbyol, line 192: This doesn't apply to Divert and it is not relevant. I would remove it

ifbyol, line 215: This wasn't tested, it would probably work, but we didn't verify the needed changes to make it work with network policies

ifbyol, line 217: We didn't do a deep research about this. it might be interesting to dedicate some time at some point to see what we can achieve.

Until then, I don't know if it worths to mention this, as we cannot guide users to know what they are looking for, and given the usage we do of Linkerd for Divert, maybe the data there is not relevant, as not all communication goes through Linkerd.

I would delete this whole subsection

ifbyol, line 248: I think it is important to mention this is only relevant for shared namespace, which are the ones where the sidecar are mandatory to work

ifbyol, line 272: We didn't explore the capabilities to do this, and it might not work. Not all the traffic is managed by Linkerd, so it might not be useful

ifbyol, line 276: I wouldn't mention ServiceProfile. It is not something we use/need to make Divert work, and we didn't test the implications of using them

src/content/core/divert.mdx
```diff
@@ -20,9 +18,74 @@ Divert transforms how teams develop microservices by making selective service de
...
+
+**Application requirements:**
+
+- A microservice architecture with multiple services
+- Services that can propagate HTTP headers to downstream calls
+- A stable, complete deployment to serve as the shared environment
+
+### Step 1: Configure the Driver (Admin)
+
+The nginx driver is enabled by default and requires no configuration. To extend routing to service-to-service calls, you can optionally install Linkerd following the [Linkerd installation guide](../self-hosted/install/divert/linkerd-installation.mdx).
```
ifbyol, line 47: I would force users to use nginx driver with Linkerd installation. Not as optional.

it might not be clear which scenarios wouldn't work without Linkerd and it might look like Divert doesn't work. I think we should enforce to make the nginx driver work with Linkerd always

src/content/core/divert.mdx
```diff
@@ -20,9 +18,74 @@ Divert transforms how teams develop microservices by making selective service de
...
+  divert:
+    driver: nginx # or 'istio' if your cluster uses the istio driver
+    namespace: staging
+```
+
+Your services must also propagate the `baggage` header to downstream calls. See [Header Propagation Requirement](#header-propagation-requirement) for code examples.
+
+### Step 4: Deploy and Verify
+
+Run `okteto deploy`. Okteto deploys your modified services, creates an ingress that injects the routing header, and gives you a unique URL such as `https://frontend-alice-myapp.okteto.example.com`.
```
ifbyol, line 82: If this is a generic description of Divert, we should not mention ingress. That only applies to nginx driver.

src/content/core/divert.mdx
ifbyol, line 97: We have flexiblity to support both cases. Your own copy/version of database or queue in case you want fully isolated data management (it is recommended in cases like working on databases migrations) I think it is important to highlight that both things are supported (your own database, that could even use volume snapshots, or the shared one)

ifbyol, line 119: Do we want to accomodate this sample to the movies-with-divert? In the latest version of movies with divert, frontend only talks with api gateway, and it is that one the one talking with the other services

ifbyol, line 282: I would say that Linkerd is required. If not, there are scenarios that wouldn't work for this driver and it would be harder to explain

src/content/core/divert.mdx
```diff
@@ -87,6 +150,129 @@ graph TB
...
+
+### Creating the Shared Namespace
+
+You have two options for creating shared namespaces:
+
+1. **Global Preview Environment** (recommended for most teams)
+   - Created via Okteto's [Preview Environment](../previews/index.mdx) system
+   - Automatically managed by Okteto
+   - Fits teams that already redeploy from CI/CD pipelines
+   - Requires: preview automation configured in your repository
```
ifbyol, line 165: This is not needed, it could be created and managed manually with `okteto preview deploy....`. It could be manual or automatic

src/content/core/divert.mdx
```diff
@@ -87,6 +150,129 @@ graph TB
...
+
+2. **Regular Namespace**
+   - Standard Okteto Namespace deployed manually or via CI
+   - More control over deployment timing
+   - Fits stable staging or demo environments
+   - Requires: manual deployment management
```
ifbyol, line 171: It can be also automatic. We have GitHub actions that can automate the creation of a namespace and pipeline deployments. This depends on what the user wants

src/content/core/divert.mdx
```diff
@@ -87,6 +150,129 @@ graph TB
...
+
+:::info Access Requirements
+All developers need **read access** to the shared namespace services for Divert routing to work. Configure namespace permissions to ensure team members can access shared resources without deployment permissions.
```
ifbyol, line 174: This depends on the the driver. For Istio driver, developers need write access for VirtualServices.

In the case of nginx, developers only need raed access, but Okteto's service account needs write access on that namespace.

src/content/core/divert.mdx
```diff
@@ -87,6 +150,129 @@ graph TB
...
+:::
+
+### Infrastructure Configuration Options
+
+When setting up your shared environment, you can configure databases and message queues in two ways:
+
+#### Option 1: Shared Infrastructure (Most Common)
```
ifbyol, line 181: I'm not sure this scenario should be the most common one. I think we should be conservatives and recommend the other case. It is easy that you end up creating some data that might affect the shared environment.

I think it is important to highlight the main cons of each approach, and explain clearly the implications

src/content/core/divert.mdx
```diff
@@ -87,6 +150,129 @@ graph TB
...
+- Schema changes are frequent
+- Working with sensitive data
+
+:::tip Hybrid Approach
+You can mix both patterns: share stable databases (e.g., user accounts) while running isolated databases for services under active development (e.g., new feature schemas).
+:::
+
+### Linkerd Configuration for Shared Namespaces
+
+If you use Linkerd with the nginx driver, sidecar injection must be enabled in both your shared namespace and developer namespaces for service-to-service routing to work:
```
ifbyol, line 243: Sidecar inject should be enabled only in shared namespaces. Developers namespace should not have the sidecar injection

ifbyol, line 243: Again, I would enforce that Linkerd is required for nginx driver

src/content/core/divert.mdx
```diff
@@ -87,6 +150,129 @@ graph TB
...
+   kubectl get pods -n staging -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.spec.containers[*].name}{"\n"}{end}'
+   ```
+   Each pod should show both application and `linkerd-proxy` containers
+
+3. **Test service connectivity:**
+   ```bash
+   kubectl exec -n staging deployment/frontend -- curl http://api-service/health
+   ```
+
+4. **Confirm ingress is accessible:**
```
ifbyol, line 271: If this section is generic about Divert, we should not mention ingress. We should probably say public endpoints

src/content/core/divert.mdx
```diff
@@ -107,18 +293,18 @@ Both drivers now use the same baggage header format: `baggage: okteto-divert=<na
 **Choose nginx driver (default) when:**
 - Using standard Okteto installation
 - You want the simplest setup
-- Optional: You can add Linkerd for advanced service mesh features
+- Optional: You can add Linkerd to extend routing to service-to-service calls
```
ifbyol, line 296: Same than before. Without this, only a basic scenario would work and we would need to enter into details about which scenarios wouldn't be supported. I think it is simpler and easier if we just say that Linkerd is required

src/content/core/divert.mdx
```diff
@@ -107,18 +293,18 @@ Both drivers now use the same baggage header format: `baggage: okteto-divert=<na
...
 
 **Choose istio driver when:**
 - Your cluster already has Istio installed
 - You prefer Istio's VirtualService-based routing
 - Your team is familiar with Istio configuration
 
 :::tip
-**Istio vs Linkerd**: These serve different purposes. **Istio is a divert driver** (alternative to nginx), while **Linkerd enhances the nginx driver** (optional add-on). You cannot use both Istio driver and Linkerd together. See the [admin configuration guide](self-hosted/install/divert/index.mdx) for details.
+**Istio vs Linkerd**: These serve different purposes. **Istio is a divert driver** (alternative to nginx), while **Linkerd extends the nginx driver** (optional add-on for service-to-service routing). You cannot use both Istio driver and Linkerd together. See the [admin configuration guide](../self-hosted/install/divert/index.mdx) for details.
```
ifbyol, line 304: I would not say that Linkerd extends the nginx driver. I would say that is required for nginx driver

src/content/core/divert.mdx
ifbyol, line 360: We can mention here that as we use a standard header `baggage`, there are a lot of libraries for a lot of languages that already take care of propagating the header, as it is a common header used for tracing or context sharing. So, no need to do your own implementation of the header propagation

ifbyol, line 406: In this sample we are missing how to extract the value of the baggage header

src/content/development/using-divert.mdx
ifbyol, line 34: It is important to mention that the header is automatically injected in the endpoints for the diverted/user's namespace.

It is also important to mention here that any public endpoint from the staging environment is also generated for the diverted/user's namespace. I mean, we don't only generate the endpoints for the services deployed, we also generate different ones for the services not deployed (if they are generated in the shared namespace)

ifbyol, line 114: Already mentioned in the general divert page, but it might also fit here. We can mention that there are libraries in a lot of programming languages as we use baggage header which is commonly used to propagate context or tracing information between services

ifbyol, line 279: We might refer to OKTETO_DIVERTED_NAMESPACE env var, as we are injecting it into diverted namespaces to be used to consume the expected messages. Some of our samples use this pattern, you can check them

ifbyol, line 282: To use this the kafka consumer has to be configured to use Consumer Groups, it might be relevant to mention it here to avoid misunderstandings. If not, this pattern wouldn't work fine

src/content/self-hosted/install/divert/linkerd-installation.mdx
codyjlandstrom: Added the two Helm changes from the handbook (`okteto-nginx.controller.podAnnotations: linkerd.io/inject: ingress` and `unsupported.linkerdDivert.enabled: true`), the Gateway API CRD prerequisite, the 1.37 / CLI 3.12 minimums, and verification commands for the okteto-nginx proxy, the Divert controller and the CRDs (85c50a2a).

One question: this is the first time an `unsupported.*` value appears in the public docs. Is that the key customers should use, or is there a plan to move the flag out of `unsupported` before we publish?

codyjlandstrom: Removed. The section now says injection is only for shared environments, that developer namespaces must not have it (the DNS fallback covers them), and gives the two ways to do it: `okteto preview deploy --label=okteto-shared`, or annotating a regular namespace before deploying.

src/content/self-hosted/install/divert/linkerd-installation.mdx
codyjlandstrom: Changed across all four Divert pages: Linkerd is now described as required for the nginx driver. The intro explains why in one sentence (without it, calls from the shared environment never reach diverted services).

src/content/self-hosted/install/divert/linkerd-installation.mdx
codyjlandstrom: Folded into a new 'Enable Sidecar Injection in Shared Environments' section that says only shared environments need it.

src/content/self-hosted/install/divert/linkerd-installation.mdx
codyjlandstrom: Removed, along with the Linkerd Viz install step and the monitoring section.

src/content/self-hosted/install/divert/linkerd-installation.mdx
codyjlandstrom: Removed the whole benefits list. The page now only says Linkerd is the service mesh that does the header-based routing for service-to-service calls, and the admin page has a note that Divert uses no other Linkerd features.

src/content/self-hosted/install/divert/linkerd-installation.mdx
codyjlandstrom: Removed.

codyjlandstrom: Removed the network policies section from this page. The generic note on the Configure Divert page (allow cross-namespace traffic) was already on main, so I left it, and it makes no claim about Linkerd.

codyjlandstrom: Deleted the subsection.

codyjlandstrom: Retitled to 'Sidecar Not Injected in the Shared Environment' and added that developer namespaces don't get sidecars.

codyjlandstrom: Removed. The troubleshooting now uses the checks from the handbook: Divert resources, HTTPRoutes in the shared namespace, Divert controller logs, and the Gateway API CRD error.

codyjlandstrom: Removed.

src/content/core/divert.mdx
codyjlandstrom: Done. Step 1 now says the nginx driver requires Linkerd and links to the install guide.

src/content/core/divert.mdx
codyjlandstrom: Changed to 'public endpoints', and added that Okteto also creates endpoints in your namespace for the shared services you didn't deploy.

src/content/core/divert.mdx
codyjlandstrom: Rewrote that bullet and the Infrastructure Configuration Options section: both patterns are supported, isolation is the safe default when the service writes or consumes from a queue, each option lists its implications, and Volume Snapshots are called out for cloning the shared database.

codyjlandstrom: Updated the diagram to match movies-with-divert: frontend → API gateway → catalog / rent, rent → PostgreSQL + Kafka, worker consuming from Kafka. Alice diverts catalog in the example so the service-to-service routing shows.

codyjlandstrom: Done. The table now says 'Requires Linkerd'.

src/content/core/divert.mdx
codyjlandstrom: Removed that requirement. It now says previews can be created manually with `okteto preview deploy` or from CI, and mentions `--label=okteto-shared` for the Linkerd annotation.

src/content/core/divert.mdx
codyjlandstrom: Changed to 'deployed manually or from CI'. Also added that a regular namespace must be shared explicitly with each developer.

src/content/core/divert.mdx
codyjlandstrom: Split by driver: nginx needs developers with read access and Okteto's service account with write access to the shared namespace; istio needs developers with write access to the VirtualServices.

src/content/core/divert.mdx
codyjlandstrom: Dropped 'Most Common'. Isolation is now the recommended default when the service writes or consumes from a queue, and each option has an Implications list.

src/content/core/divert.mdx
codyjlandstrom: Fixed. Shared namespace only, developer namespaces must not have injection, and the wrong 'inherit' line is gone.

codyjlandstrom: Done.

src/content/core/divert.mdx
codyjlandstrom: Changed to 'public endpoint'.

src/content/core/divert.mdx
codyjlandstrom: Done.

src/content/core/divert.mdx
codyjlandstrom: Done.

src/content/core/divert.mdx
codyjlandstrom: Added a paragraph saying `baggage` is the W3C context propagation header and that OpenTelemetry and most tracing libraries propagate it automatically, so most teams don't need custom code. Same note on the Using Divert page.

codyjlandstrom: Added the `@RequestHeader("baggage")` extraction to the Java sample.

src/content/development/using-divert.mdx
codyjlandstrom: Added both points to the 'when you run okteto deploy' list: the header is injected on every endpoint in your namespace, and Okteto copies every shared endpoint you didn't deploy into your namespace, pointing back to the shared services. Also listed the DNS fallback and the two injected env vars.

codyjlandstrom: Added.

codyjlandstrom: Rewrote the SQS and Kafka samples around `OKTETO_DIVERTED_ENVIRONMENT` (the CLI constant and the 1.3x release notes use `_ENVIRONMENT`, not `_NAMESPACE`). The producer forwards the `baggage` value, and the consumer compares the `okteto-divert` key with the env var, which is unset in the shared namespace. Linked the movies-with-divert worker as the reference implementation.

codyjlandstrom: Added a paragraph saying the pattern only works with consumer groups and a per-environment group ID, and why (with a shared group ID each message reaches only one environment). The sample now derives the group ID from the environment.

PR #1197, opened 2026-04-13
Clarify Istio divert virtualServices and hosts field descriptions

codyjlandstrom:
The existing docs showed YAML syntax for these fields but didn't explain what each one does or how they differ. Added descriptions clarifying that virtualServices modifies routing in the shared namespace while hosts creates dedicated copies in the developer namespace.


src/content/reference/okteto-manifest.mdx
```diff
@@ -383,8 +383,8 @@ deploy:
 ```
 
 - `driver`: Specifies the backend to divert requests. Use `istio` to use the Istio driver.
-- `virtualServices`: A list of virtual services to divert. Each virtual service is defined by it's name, Namespace, and an optional list of routes to be diverted. By default, all routes are diverted.
-- `hosts`: The list of hosts you want to divert in the developer namespace. Requests to these virtual services will have the key `okteto-divert` as part of the `baggage` header injected.
+- `virtualServices`: A list of Istio VirtualService resources in the shared namespace to modify for traffic diversion. Okteto adds header-based routing logic to these virtual services so that requests with the baggage header are sent to the developer's namespace. Each entry is defined by its name, namespace, and an optional list of routes to be diverted. By default, all routes are diverted.
```
ifbyol:
```suggestion
- `virtualServices`: A list of Istio VirtualService resources in the shared namespace to modify for traffic diversion. Okteto adds header-based routing logic to these virtual services so that requests with the baggage header are sent to the developer's namespace. Each entry is defined by its name, namespace, and an optional list of routes to be diverted.
```

In fact, you need to specify at least one virtual service, if not, it would fail. I know it is not related with the changes you are doing, but I saw it this morning while I was gathering context to answer some questions about it

ifbyol (approved):

src/content/reference/okteto-manifest.mdx
codyjlandstrom: Ahh thank you for this clarification

src/content/development/using-divert.mdx
```diff
@@ -53,6 +53,12 @@ deploy:
...
+
+- **`hosts`**: Creates a dedicated copy of the listed virtual services in your development namespace, each with its own host (e.g., `https://service-a-<dev-namespace>.<subdomain>`). Requests reaching this host automatically have the baggage header injected. Internally, the copied virtual service still points to the original virtual service in the shared namespace, ensuring the rest of the request flow goes through the shared environment with the divert header applied. This is useful for services you are **not** deploying as part of your development environment — it gives you a dedicated URL to reach your version of the app without having to manually use the staging endpoint with the baggage header. Services you deploy in your own namespace already have their own endpoints.
```
rberrelleza: can we mention that this is optional?

PR #1211, opened 2026-04-22
Fix installer CodeBlock: runner indented under installer (all versions)

rberrelleza:
## Summary

- Supersedes #1210 — the previous fix was insufficient
- The root cause: having the template literal on a separate line from `<CodeBlock>` creates a whitespace text node (`\n  `) that gets prepended to the code content, making `installer:` render with 2 spaces of indentation — the same as `runner:`, making them appear at the same level
- Fix: move the template literal to open directly on the `<CodeBlock>` tag line, eliminating any whitespace text node entirely

**Before:**
```jsx
<CodeBlock language="yaml">
  {`installer:
  runner:
    ...`}
</CodeBlock>
```

**After:**
```jsx
<CodeBlock language="yaml">{`installer:
  runner:
    ...`}</CodeBlock>
```

## Affected versions

All versions: 1.32–1.43 and `src/content` (next release).

## Test plan

- [x] Verify the `installer` code block on the helm configuration page shows `runner` correctly indented under `installer`
- [x] Spot-check across a few versioned docs pages

### DIFF
--- a/src/content/self-hosted/helm-configuration.mdx
+++ b/src/content/self-hosted/helm-configuration.mdx
@@ -516,23 +516,24 @@ The jobs that deploy your [development environments from Git](development/deploy
 - `securityContext`: The security context for the installer job container. It's not set by default, and it follows the [same syntax used in Kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/)
 - `extraInitContainers`: List of init containers to add to the installer job pods.
 
-<CodeBlock language="yaml">
-  {`installer:
-  runner:
-    repository: okteto/pipeline-runner
-    tag: ${variables.chartVersion}
-  extraEnv:
-  - name: NO_PROXY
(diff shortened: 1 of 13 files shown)


codyjlandstrom (changes requested):
Since we already merged the dependency upgrade, I think we should cherry pick the codeblock changes from this.

rberrelleza:
@codyjlandstrom problem is that the dependency upgrade was merged into this branch, not into main. 😩

codyjlandstrom:
> @codyjlandstrom problem is that the dependency upgrade was merged into this branch, not into main. 😩

In that case, we should cherry pick the changes we want to keep into a new PR.

rberrelleza:
> > @codyjlandstrom problem is that the dependency upgrade was merged into this branch, not into main. 😩
> 
> In that case, we should cherry pick the changes we want to keep into a new PR.

done!

PR #1220, opened 2026-04-22
docs: add ARM support documentation

codyjlandstrom:
## What

Adds a dedicated ARM support page and updates existing installation guides to reference it.

**New file:** `src/content/self-hosted/manage/arm-support.mdx`

**Updated files:**
- `src/content/get-started/install/google-gke.mdx` — adds a link from the existing ARM toleration warning to the new guide
- `src/content/get-started/install/amazon-eks.mdx` — adds a Beta ARM note in the cluster creation section
- `src/content/self-hosted/index.mdx` — adds ARM Support to the index

## Why

ARM support is shipping in Phase 1 (Q3 2025). Documentation should be available at GA/Beta declaration so customers have a clear reference for what's supported, how to configure it, and what's out of scope.

Source: [ARM Proposal](https://www.notion.so/okteto/ARM-Proposal-254166deb6a080838dc6ddcc1b663fda)

## Content summary

- GCP Tau T2A (`t2a-standard-4`) — **GA**, backed by automated CI on ARM clusters every release
- AWS Graviton 2/3 (`m7g.xlarge`) — **Beta**, validated manually on new Kubernetes versions
- Helm toleration config for GKE ARM node taints (already partially documented; now consolidated)
- Full limitations table (no mixed clusters, no migration from x86, no BYOC, no Divert, etc.)
- Dependency ARM compatibility note (Redis, Reloader, ingress-nginx)

## Open questions for Engineering review

Before marking this ready, please confirm:

- [ ] **EKS tolerations** — Does EKS auto-taint Graviton nodes with `kubernetes.io/arch=arm64:NoSchedule` like GKE does? If yes, add the toleration block to the EKS section (same pattern as GKE).
- [ ] **EKS AMI family** — Is `amiFamily: AmazonLinux2023` correct for Graviton, or is a different AMI family needed?
- [ ] **GKE cluster creation command** — Validate that the `gcloud` command in the ARM guide matches what was actually used for provisioning (zone, Standard vs Autopilot, any other flags).
- [ ] **Kubernetes version range** — Should we document a minimum supported K8s version for ARM clusters? CI runs `1.35arm`; is there a lower bound?
- [ ] **GA declaration** — Is GCP ARM officially declared GA, or still pending? If still pending, the status table should say "GA (pending)" until the announcement.

### DIFF
--- a/src/content/self-hosted/manage/arm-support.mdx
+++ b/src/content/self-hosted/manage/arm-support.mdx
@@ -1,0 +1,162 @@
+---
+title: ARM Support
+description: How to install and configure Okteto on ARM-based Kubernetes clusters
+sidebar_label: ARM Support
+id: arm-support
+---
+
+Okteto supports installation on ARM-based Kubernetes clusters for self-hosted deployments. This lets you run the Okteto control plane on ARM infrastructure, which can offer better cost efficiency compared to equivalent x86 instance types.
+
+## Support Status
+
+| Cloud Provider | Recommended Instance | Status |
+|---|---|---|
+| Google GKE | Tau T2A (`t2a-standard-4`) | GA |
+| Amazon EKS | Graviton 2/3 (`m7g.xlarge`) | Beta |
+
+ARM support is covered by the same release pipeline as x86. All control plane images are published as multi-arch (amd64 + arm64).
+
+:::note
+ARM support is available for **self-hosted deployments only**. BYOC and SaaS deployments are not supported.
+:::
+
+## Requirements
+
+- **Fresh installations only.** Migrating an existing x86 cluster to ARM is not supported. ARM clusters must be provisioned from scratch.
+- **Homogeneous clusters.** All nodes in the cluster must be ARM-based. Mixed architecture clusters (ARM + x86 nodes) are not supported.
+
+## Google GKE (GA)
+
+ARM support on GCP using Tau T2A instances is generally available. Automated tests run against ARM clusters on GCP for every release and in weekly scheduled runs.
+
+### Cluster Configuration
+
+We recommend creating a GKE cluster using Tau T2A nodes. The recommended instance type is `t2a-standard-4`:
+
+```bash
+gcloud container clusters create CLUSTER_NAME \
+  --zone ZONE \
+  --machine-type=t2a-standard-4 \
+  --num-nodes=3 \
+  --disk-size=250
+```
+
+:::note
+Tau T2A instances are only available in certain GCP regions and zones. See the [GCP documentation](https://cloud.google.com/compute/docs/general-purpose-machines#t2a_machines) for regional availability.
+:::
+
+### Helm Configuration
+
+GKE automatically applies `kubernetes.io/arch=arm64:NoSchedule` taints to ARM nodes. You must add the following tolerations to your `config.yaml` so that Okteto components can be scheduled on those nodes:
+
+```yaml
+globals:
+  tolerations:
+    okteto:
+      - key: "kubernetes.io/arch"
+        operator: "Equal"
+        value: "arm64"
+        effect: "NoSchedule"
+    dev:
+      - key: "kubernetes.io/arch"
+        operator: "Equal"
+        value: "arm64"
+        effect: "NoSchedule"
+ingress-nginx:
+  controller:
+    tolerations:
+      - key: "kubernetes.io/arch"
+        operator: "Equal"
+        value: "arm64"
+        effect: "NoSchedule"
+okteto-nginx:
+  controller:
+    tolerations:
+      - key: "kubernetes.io/arch"
+        operator: "Equal"
+        value: "arm64"
+        effect: "NoSchedule"
+reloader:
+  reloader:
+    deployment:
+      tolerations:
+        - key: "kubernetes.io/arch"
+          operator: "Equal"
+          value: "arm64"
+          effect: "NoSchedule"
+```
+
+For the full installation walkthrough, follow the [GKE installation guide](get-started/install/google-gke.mdx).
+
+## Amazon EKS (Beta)
+
+:::warning Beta
+ARM support on Amazon EKS is in **Beta**. Validation is performed manually when introducing support for new Kubernetes versions. We recommend testing in a non-production environment before using this in production.
+:::
+
+AWS Graviton 2 and Graviton 3 instances are supported. The recommended instance type is `m7g.xlarge`, which is equivalent in size to the standard `m5.xlarge` but typically delivers better performance at a lower cost.
+
+### Cluster Configuration
+
+```bash
+eksctl create cluster -f - <<EOF
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+
+metadata:
+  name: ${CLUSTER_NAME}
+  region: ${AWS_REGION}
+  version: "${K8S_VERSION}"
+
+autoModeConfig:
+  enabled: false
+
+iam:
+  withOIDC: true
+
+nodeGroups:
+  - name: ng-1
+    instanceType: m7g.xlarge
+    desiredCapacity: 3
+    volumeSize: 250
+    volumeType: gp3
+    amiFamily: AmazonLinux2023
+EOF
+```
+
+### Helm Configuration
+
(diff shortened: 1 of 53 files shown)


src/content/self-hosted/manage/arm-support.mdx
```diff
@@ -0,0 +1,162 @@
+---
+title: ARM Support
+description: How to install and configure Okteto on ARM-based Kubernetes clusters
+sidebar_label: ARM Support
+id: arm-support
+---
+
+Okteto supports installation on ARM-based Kubernetes clusters for self-hosted deployments. This lets you run the Okteto control plane on ARM infrastructure, which can offer better cost efficiency compared to equivalent x86 instance types.
```
irespaldiza: I’m not sure we should position cost efficiency as the reason for supporting ARM. While it can be a side benefit, the main goal here is to enable users to run development environments that match their production architecture.

If an application runs on ARM in production, it should also be developed and tested on ARM to avoid subtle compatibility issues and ensure consistency in testing and debugging.

Framing this around cost might unintentionally encourage mixing ARM and x86 environments, which is generally not a good practice.

src/content/self-hosted/manage/arm-support.mdx
```diff
@@ -0,0 +1,162 @@
...
+          value: "arm64"
+          effect: "NoSchedule"
+```
+
+For the full installation walkthrough, follow the [GKE installation guide](get-started/install/google-gke.mdx).
+
+## Amazon EKS (Beta)
+
+:::warning Beta
+ARM support on Amazon EKS is in **Beta**. Validation is performed manually when introducing support for new Kubernetes versions. We recommend testing in a non-production environment before using this in production.
```
irespaldiza: I think this sentence is ambiguous for two reasons.

On one hand, I would avoid using the term “production”. While it can be technically correct from a certain perspective, an Okteto cluster is still a development and testing environment. I would personally suggest something like “operational cluster” to reduce ambiguity, although this is just a suggestion and not a required change.

Regarding testing installations or upgrades, this is something we recommend in all cases, and I do not think it should be associated specifically with a beta state. The real point is that while the functionality itself is complete, the testing framework is not yet mature, so there may still be bugs or unexpected issues.

Because of this, the message should not imply that customers are responsible for testing the software themselves, but rather that the feature does not yet have the expected level of maturity, and that using it in operational environments involves some level of risk.

src/content/self-hosted/manage/arm-support.mdx
```diff
@@ -0,0 +1,162 @@
...
+:::
+
+AWS Graviton 2 and Graviton 3 instances are supported. The recommended instance type is `m7g.xlarge`, which is equivalent in size to the standard `m5.xlarge` but typically delivers better performance at a lower cost.
```
irespaldiza: As mentioned in a previous comment, I would position ARM support as a way to enable development environments that accurately reflect production. I would avoid linking it to cost or performance considerations, since those are not necessarily true in every case and are outside of Okteto’s control.

src/content/self-hosted/manage/arm-support.mdx
```diff
@@ -0,0 +1,162 @@
...
+    desiredCapacity: 3
+    volumeSize: 250
+    volumeType: gp3
+    amiFamily: AmazonLinux2023
+EOF
+```
+
+### Helm Configuration
+
+<!-- TODO: Engineering to confirm whether EKS ARM nodes require tolerations (similar to GKE) and provide exact values if so -->
```
irespaldiza: As of today, AWS does not add taints that require any additional configuration.

src/content/self-hosted/manage/arm-support.mdx
```diff
@@ -0,0 +1,162 @@
...
+|---|---|
+| Migration from x86 clusters | No migration tooling; fresh install required |
+| Mixed architecture clusters | All nodes must be the same architecture |
+| BYOC deployments | Not supported |
+| Azure AKS ARM | Not supported |
+| GCP Axion (C4A) instances | Not supported |
+| Oracle Ampere A1/A2 | Not supported |
+| Bare metal ARM | Not supported |
+| Divert | Not supported on ARM (planned for a future release) |
+| External Secrets Operator | Not supported on ARM |
```
irespaldiza: External Secrets Operator is not part of Okteto, so I do not understand this reference.

irespaldiza (changes requested):

src/content/self-hosted/manage/arm-support.mdx
```diff
@@ -0,0 +1,162 @@
...
+| GPU support | Not supported on ARM |
```
irespaldiza: This limitation comes from AWS Graviton instances. As written, it may give the impression that it is a limitation of Okteto, which is not the case.

codyjlandstrom:
### Code review

Found 2 issues (focused on items not already raised in @irespaldiza's review):

1. `self-hosted` should be `Self-Hosted` (CLAUDE.md says "`Self-Hosted` | Always capitalized, hyphenated"). Two occurrences in the new file:

https://github.com/okteto/docs/blob/5f43c3f1ce8a7793fb775e0e3962c34b61defb85/src/content/self-hosted/manage/arm-support.mdx#L7-L9

https://github.com/okteto/docs/blob/5f43c3f1ce8a7793fb775e0e3962c34b61defb85/src/content/self-hosted/manage/arm-support.mdx#L19-L21

2. First reference to `BYOC` should be spelled out as `Bring Your Own Cloud (BYOC)` ([CLAUDE.md](https://github.com/okteto/docs/blob/main/CLAUDE.md) says "`BYOC` is always uppercase. First reference: `Bring Your Own Cloud (BYOC)`"). The page uses `BYOC` at line 20 and again in the Limitations table without ever spelling it out:

https://github.com/okteto/docs/blob/5f43c3f1ce8a7793fb775e0e3962c34b61defb85/src/content/self-hosted/manage/arm-support.mdx#L19-L21

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/content/self-hosted/manage/arm-support.mdx
codyjlandstrom: Good point. I reframed the opening around dev/prod parity instead of cost.
> Okteto supports installation on ARM-based Kubernetes clusters for Self-Hosted deployments. Running Okteto on ARM lets developers build and test against the same architecture as their target operational clusters, avoiding compatibility differences that can surface when developing on a different architecture than the one that runs the application.

Borrowed "operational clusters" from your line 94 suggestion since it fits here too.

src/content/self-hosted/manage/arm-support.mdx
codyjlandstrom: Reworked the admonition to drop "production" and shift the framing from "customer should test this" to "test coverage is still maturing". 

> ARM support on Amazon EKS is in Beta. The functionality is complete, but automated test coverage on EKS ARM is still maturing — validation is performed manually when introducing support for new Kubernetes versions. Until coverage matures, running this on operational clusters carries a higher chance of encountering an unexpected issue.

Let me know if "carries a higher chance of encountering an unexpected issue" is still too close to putting the burden on the customer?

src/content/self-hosted/manage/arm-support.mdx
codyjlandstrom: Dropped the "better performance at a lower cost" claim. The sentence now reads:

> AWS Graviton 2 and Graviton 3 instances are supported. The recommended instance type is m7g.xlarge, which is equivalent in size to the standard m5.xlarge used in the default EKS guide.

src/content/self-hosted/manage/arm-support.mdx
codyjlandstrom: Resolved the TODO with your confirmed answer in:

> EKS does not apply architecture taints to Graviton nodes, so no additional tolerations are required for ARM scheduling.

src/content/self-hosted/manage/arm-support.mdx
codyjlandstrom: Removed

irespaldiza (approved):

src/content/self-hosted/manage/arm-support.mdx
codyjlandstrom: Good call. Made a quick tweak

PR #1263, opened 2026-06-11
docs(agentic): restructure section around install-first landing page

codyjlandstrom:
## Summary

Restructures the Agentic Workflows section so readers can act sooner, modeled on the install-first shape of [Clerk's skills docs](https://clerk.com/docs/guides/ai/skills). The section had ~3,000 words across five pages with heavy duplication, and the install command sat at the bottom of the landing page.

- **Merge the Claude Code Plugin page into the section index.** The landing page now runs: value prop → install commands → a "Skills and commands" inventory → example prompts → mode comparison. The two pages overlapped ~60%.
- **Document the `okteto-onboarding` skill.** The plugin ships it, but the docs never mentioned it. It now has a row in the skills table ("the project has no `okteto.yaml` and you ask to set it up for Okteto").
- **De-duplicate.** The `okteto up` rule was explained five times across the section; it's now one canonical explanation in Best Practices plus one-line pointers. The per-page CLI permission tables on the workflow pages are replaced with links to the canonical Command rules table.
- **Redirect.** `/docs/agentic/claude-code-plugin` 301s to `/docs/agentic` via netlify.toml (non-forced, so it only applies once the published version no longer has the page).

## Validation

- `yarn build` passes with no broken links or anchors
- Verified rendering locally (sidebar, tables, TOC) on the dev server
- Checked against STYLE_GUIDE.md: no banned words, noun-phrase headings, inline links, functional frontmatter descriptions


src/content/agentic/best-practices.mdx
```diff
@@ -74,7 +92,7 @@ okteto deploy
 
 ### Agent hangs and becomes unresponsive
 
-The most common cause is the agent running `okteto up`, which is interactive and sits waiting for terminal input. Stop the agent, and switch to `okteto exec` (collaborative) or `okteto deploy` + `okteto build` (autonomous). The [Claude Code Plugin](agentic/claude-code-plugin.mdx) teaches agents to avoid this automatically.
+The most common cause is the agent running `okteto up`, which is interactive and sits waiting for terminal input. Stop the agent, and switch to `okteto exec` (collaborative) or `okteto deploy` + `okteto build` (autonomous). The [Okteto plugin for Claude Code](agentic/index.mdx#installing-the-plugin) teaches agents to avoid this automatically.
```
ifbyol: Okteto exec needs a running `okteto up` session, so this `The most common cause is the agent running `okteto up`, which is interactive and sits waiting for terminal input. Stop the agent, and switch to `okteto exec` (collaborative)` is a bit odd if nobody runs okteto up. I don't know if the expectation is that `okteto up` keeps running despite you stop the agent (which I think is the idea). Maybe we should mention explicitly that you should own the `okteto up` session while the agent uses `okteto exec`

src/content/agentic/collaborative-workflows.mdx
```diff
@@ -50,41 +48,15 @@ okteto exec -- npm test
...
-
-For full flag details on any command, see the [CLI Reference](reference/okteto-cli.mdx).
-
-### Commands agents must not run
-
-| Command | Why |
-|---------|-----|
-| `okteto up` | Interactive. Requires a human terminal and will hang if run by an agent. |
-| `okteto destroy` | Tears down the entire environment. Should only be run by the developer or with explicit policy. |
+The agent's day-to-day commands in this mode are `okteto exec`, `okteto logs`, and `okteto test`. The full list of commands agents may and may not run is in [Command rules](agentic/best-practices.mdx#command-rules). For flag details on any command, see the [CLI Reference](reference/okteto-cli.mdx).
```
ifbyol: Why `okteto logs`? If the service is in dev mode, `okteto logs` doesn't show meaningful logs from the okteto up session, those are only accesible in the terminal where the session is running

src/content/agentic/best-practices.mdx
codyjlandstrom: Good point!
Reworded: stop the agent, but in collaborative mode you keep the okteto up session running in your own terminal and the agent works against it with okteto exec; autonomous mode uses okteto deploy + okteto build.

src/content/agentic/collaborative-workflows.mdx
codyjlandstrom: for a service in dev mode the app output goes to the `okteto up` terminal, not `okteto logs`. Dropped `okteto logs` from the day-to-day list and the bug example, and added a note that live output appears in your `okteto up` terminal while the agent reproduces/verifies via `okteto exec`. `okteto logs <service>` now scoped to the other (non-dev) services in the environment.

ifbyol (approved):

PR #1293, opened 2026-07-10
refactor(tutorials): drive card logos from front matter, not title keywords

rberrelleza:
## What & why

The tutorial card grid (`src/theme/CardsList`) picked each logo by scanning the tutorial **title** for a hardcoded keyword (`aws`, `docker`, `webpack`, `launchdarkly`, `okteto`). The code even said so:

> _Unfortunately, at the moment we don't have access to custom frontmatter value. […] We'll check if the title string includes a keyword._

That workaround predates our current Docusaurus. On 3.x, custom front matter flows to sidebar items via [`sidebar_custom_props`](https://docusaurus.io/docs/sidebar/items#customizing-sidebar-items) → `useCurrentSidebarCategory().items[].customProps`. So the component now reads `customProps.logo` and the title-keyword scan is gone.

This allows tutorials to pick their own icon rather than depending on the title.

## Changes
- **`CardsList`**: read `item.customProps?.logo`; build the URL from it. Removed `findKeywordInString` + the keyword list.
- **9 tutorials**: added `sidebar_custom_props: { logo: ... }`. Files that already carried an (unused) `logo:` key now use `sidebar_custom_props`.
- **Bug fix**: the *Use Volume Snapshots* card had `logo: okteto` in front matter but its title contains no keyword, so the old scan returned nothing and it rendered **no logo**. It now shows the Okteto logo as intended.
- **React**: hoisted the docs-version lookup out of the render loop, so we no longer call a hook (`useDocById`) inside `.map()`.
- **`claude.svg`**: added so a future Claude Code tutorial can opt in with `logo: "claude"` (the original ask). ⚠️ It's a hand-drawn sunburst approximation in Anthropic clay `#D97757` — swap for the official asset if brand accuracy matters.

## Adding a logo is now trivial
1. Drop `<name>.svg` in `static/img/logos/`.
2. In the tutorial front matter: `sidebar_custom_props: { logo: "<name>" }`.

No component edits, no keyword list.

## ⚠️ One intentional behavior change: Reference cards
`CardsList` is **also** used by the References landing page. The old keyword scan *accidentally* put a logo on 3 reference cards (`Okteto CLI`, `Okteto Manifest` → okteto; `Docker Compose` → docker). Since logos were only ever meant for tutorials (and `okteto.svg` on "Okteto CLI" is redundant), I scoped logos to tutorials and **dropped the accidental reference icons**. This keeps behavior consistent across all doc versions without editing the frozen `versioned_docs/` snapshots.

If you'd rather keep reference logos: add `sidebar_custom_props: { logo: "..." }` to those reference docs (and to their versioned copies).

## Verification
- Full `docusaurus build` passes.
- Built HTML: all 9 tutorial cards render exactly one correct logo (incl. the fixed Volume Snapshots card); reference landing pages render none, consistently across 1.46 and 1.47.

### DIFF
--- a/src/theme/CardsList/index.js
+++ b/src/theme/CardsList/index.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { useDocById, useCurrentSidebarCategory } from '@docusaurus/plugin-content-docs/client';
+import { useDocsVersion, useCurrentSidebarCategory } from '@docusaurus/plugin-content-docs/client';
 import useBaseUrl from '@docusaurus/useBaseUrl';
 
 import Card from '../Card';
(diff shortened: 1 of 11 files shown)


codyjlandstrom (approved):
Nice!

PR #511, opened 2023-10-10
Document priorityClassName

AgustinRamiroDiaz:
Implemented in https://github.com/okteto/app/pull/7309
Some missing configurations I didn't add due to their sections not being present:
- installerChecker
- nodeReadinessChecker
- migration
- namespaceDestroyAll

Should I create their sections?

PR #1307, opened 2026-07-21
docs: document migration, installerChecker, and namespaceDestroyAll jobs

### LINKED ISSUE #513: Document `installerChecker`, `migration` and `namespaceDestroyAll`
[Context](https://github.com/okteto/docs/pull/511#pullrequestreview-1669022006)

This 3 sections need to be documented:
- [ ] `migration` (top priority)
- [ ] `installerChecker`
- [ ] `namespaceDestroyAll`

rberrelleza:
Closes #513

Documents the three Kubernetes jobs Okteto runs in the `okteto` namespace (`migration`, `installerChecker`, `namespaceDestroyAll`) in the Helm configuration reference.

## What changed
- **`helm-configuration.mdx`**: adds `installerChecker`, `migration`, and `namespaceDestroyAll` config sections under **Okteto Components**, giving `priorityClassName` and the other values a documented home (the gap raised in the #511 review). Each section describes the job and its configurable values.
- **`troubleshooting.mdx`**: notes that the `installerChecker` CronJob clears pipelines stuck in "progressing", linking to its Helm configuration section.

Self-contained; no dependency on other PRs.


src/content/self-hosted/helm-configuration.mdx
```diff
@@ -536,6 +536,56 @@ The jobs that deploy your [development environments from Git](development/deploy
...
+- `schedule`: The cron schedule for the checker. Defaults to `*/5 * * * *` (every 5 minutes).
+- `priorityClassName`: The priority class for the installer checker pods. The PriorityClass must already exist in your cluster before using this setting. This value has precedence over `globals.priorityClassName` if both are set. If this value is not set, the pods will inherit the priority class defined by the value set in `globals.priorityClassName`.
+
+```yaml
+installerChecker:
+  enabled: true
+  schedule: "*/5 * * * *"
+```
+
+### migration
```
ifbyol, line 553: I have doubts about documenting this. I'm in the fence. At least, if we document it, we have to be clear that the migration is important, and I would even say mandatory for Okteto, so it shouldn't be disabled. The setting to disable is there for historical reasons, but nobody should change it. The rest of properties are fine to be documented. They might be important in some scenarios

ifbyol (approved):
Approving it, but it is important to add a note in the `migration` section, or not documented `enabled` property

src/content/self-hosted/helm-configuration.mdx
codyjlandstrom: Nacho and I discussed separately - Removing for now would be the safer option and we'll look to delete this flag in the chart

rberrelleza: @codyjlandstrom can we remove the “enable” keywords in that case but leave all the other configuration?

codyjlandstrom: Yes, that works. Nacho's concern was only the `enabled` flag, and he was fine documenting the rest. I'll restore the migration section with `annotations`, `labels`, `resources`, and `priorityClassName`, drop `enabled` from the list and the example, and add a note that the job is required and runs automatically on every install and upgrade.

One thing to confirm: `installerChecker.enabled` stays documented, right? That one looks like a real toggle, unlike the migration one.

PR #1325, opened 2026-08-11
docs: strip tier badges from llms.txt link titles

codyjlandstrom:
## What

Link titles in the generated `build/docs/llms.txt` concatenated tier-badge text, e.g. `- [CatalogScaleEnterpriseSelf-Hosted](/docs/admin/catalog.md)`.

The llms-txt plugin extracts each page title from the first h1 of the raw HTML AST **before** the rehype pipeline runs, so the `TiersList` badge markup rendered inside the h1 leaks into the title — and rehype-based stripping (like `rehypeStripTierBadges`) can't reach it.

## How

Extends the existing patch-package patch for `@signalwire/docusaurus-plugin-llms-txt` so `extractFirstH1` drops `div.TiersList` elements before reading the h1 text, using the same matcher as the `rehypeStripTierBadges` rehype plugin. Fallback behavior (plain h1, `<title>` tag) is unchanged.

## Verification

- Unit-level check against the installed plugin: badge-laden h1 extracted as `CatalogScaleEnterpriseSelf-Hosted` before the patch, `Catalog` after; plain-h1 and title-tag-fallback cases unchanged.
- Full `yarn build`: all nine badge-bearing pages (`admin/catalog`, `admin/cleanup`, `admin/dashboard`, `admin/previews`, `admin/private-repositories/ssh-key`, `development/deploy/deploy-from-catalog`, `get-started/install/openshift`, `reference/okteto-cli`, `self-hosted/manage/air-gapped`) now render clean titles in `build/docs/llms.txt`, e.g. `- [Catalog](/docs/admin/catalog.md)`.

## Notes

- Complementary to the `rehypeStripTierBadges` fix (branch `claude/keen-chaum-b4a4e4`), which cleans badge text out of headings **inside** the generated Markdown twins / `llms-full.txt`. The two changes touch different files and merge independently; both are needed for fully clean output.
- The patch works around an upstream design limitation (title extraction runs before `content.beforeDefaultRehypePlugins`); worth an upstream issue proposing post-pipeline title extraction or a custom title extractor option.

### DIFF
--- a/patches/@signalwire+docusaurus-plugin-llms-txt+1.2.2.patch
+++ b/patches/@signalwire+docusaurus-plugin-llms-txt+1.2.2.patch
@@ -1,3 +1,51 @@
+diff --git a/node_modules/@signalwire/docusaurus-plugin-llms-txt/lib/transformation/title-extractor.js b/node_modules/@signalwire/docusaurus-plugin-llms-txt/lib/transformation/title-extractor.js
+index 57fd4af..82776fa 100644
+--- a/node_modules/@signalwire/docusaurus-plugin-llms-txt/lib/transformation/title-extractor.js
++++ b/node_modules/@signalwire/docusaurus-plugin-llms-txt/lib/transformation/title-extractor.js
+@@ -2,15 +2,41 @@
+  * Title extraction strategies
+  * Multiple strategies for extracting document titles from HTML
+  */
++import { select } from 'hast-util-select';
++import { toString } from 'hast-util-to-string';
+ import { DEFAULT_DOCUMENT_TITLE, HTML_SELECTORS } from '../constants';
+ import { selectText } from '../utils/html';
++/**
++ * Okteto's TiersList tier badges (`<div class="TiersList">…`) render inside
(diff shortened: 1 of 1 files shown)


ifbyol (approved):

codyjlandstrom:
The deploy preview failure is fixed — all checks are green on be7094e.

Root cause: Netlify restores `node_modules` from its build cache with the previously applied patch-package patch baked in. This PR edited that patch file, so `patch-package` in postinstall could neither apply it (the `url.js` hunks were already applied in the cache) nor detect it as already applied (the new title-extractor hunks weren't), and the deploy failed in seconds — while fresh local installs built fine.

Fix: split the patch into patch-package sequenced files — `001+md-link-paths.patch` (byte-identical to the existing patch on main, so cached node_modules pass the already-applied check) and `002+strip-tier-badges-from-titles.patch` (applies cleanly to the untouched cached file). Verified locally against fresh, cached-stale, and fully patched node_modules states. Future patch changes should be added as new sequence files rather than edits to existing ones, to keep cached builds working.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1328, opened 2026-08-13
Document the ingress-nginx default certificate override

rberrelleza:
## Problem

Setting `wildcardCertificate.create: false` alone is not enough to switch Okteto from its built-in self-signed wildcard certificate to a cert-manager / Let's Encrypt one.

The `ingress-nginx` controller reads its default certificate from `ingress-nginx.controller.extraArgs.default-ssl-certificate`, which is not derived from `wildcardCertificate.name`. If you only set `create: false`, the controller keeps serving its previous default certificate. It fails silently: cert-manager reports the certificate as `Ready`, but browsers still show a warning.

## What the docs said

The three community guides linked from the cert-manager page already include both settings, and so does the [bring your own certificate](https://www.okteto.com/docs/self-hosted/install/certificates/bring-your-own-certificate/) page. Two gaps remained:

- The [cert-manager page](https://www.okteto.com/docs/self-hosted/install/certificates/cert-manager/) has no Helm configuration of its own. It links out to the community guides and shows a video, so nothing on the page itself tells you about the override.
- The [Helm configuration reference](https://www.okteto.com/docs/self-hosted/helm-configuration/) documents `create` ("If set to false, Okteto will not create a self-signed wildcard certificate") and `name`, with no mention of the ingress-nginx argument. Anyone configuring from the reference alone hits the silent failure.

## Changes

- Add a "Configure Okteto to use your certificate" section to the cert-manager page with the `wildcardCertificate` + `ingress-nginx.controller.extraArgs` pair, mirroring the bring-your-own-certificate page, plus a warning that both settings are required.
- Add a note under `wildcardCertificate` in the Helm configuration reference.

Applied to both the next version and 1.47.

## Open questions for reviewers

- The community guides were last updated between 2023-01 and 2024-10. They may be worth a refresh, or worth folding into the docs so the cert-manager page is self-contained.
- The guides' `Issuer` lists an `http01` solver before `dns01`, while this page says you must use DNS01 for a wildcard. Minor, but confusing.
- A chart-side validation that fails the install when `create: false` and `default-ssl-certificate` still points at the self-signed secret would close this properly.


src/content/self-hosted/install/certificates/cert-manager.mdx
```diff
@@ -17,6 +17,27 @@ The Okteto community maintains guides on using cert-manager with different cloud
 - [Google Cloud DNS](https://community.okteto.com/t/how-do-i-configure-okteto-with-cert-manager-and-google-cloud-dns/274/2)
 - [Azure Cloud DNS](https://community.okteto.com/t/how-do-i-configure-okteto-with-cert-manager-and-azure-cloud-dns/275/2)
 
+## Configure Okteto to use your certificate
```
ifbyol, line 20: In all the 3 guides above we already indicate the helm setting to be used to configure your certificate in Okteto. If we are guide them to those guides, and even the new section refers to them, why do we want to add it here too? Shouldn't we keep that information only in 1 place (community guides or public documentation)

rberrelleza: This has been a source of constant confusion for customers. I think it's better to be explicit and repetitive rather than require customers to dig through multiple documents, especially because this is a non-trivial configuration.

ifbyol: But then we should consider to move all the documents. This is not solving the problem of making customers dig through multiple documents. They still need to go to those documents to know what they have to do based on the provider.

Duplicating the configuration will make that at some point, both sources could diverge, mainly because the source of them are different (they are not 2 pages within the same repository). That divergence would be more confusing that always sending them to the community post where steps and configuration are documented

rberrelleza: I think this helps in several ways:

1. For customers that know how to use cert-manager, they still need the okteto-specific configuration. This provides it. I think this is specific enough to live in our docs.

2. When using agents to configure, this is more effective as it reduces the hops.

3. The community guides are not maintained nor monitored by definition (it's a community resource, not a canonical source of info). in the scenario you describe I think we are more likely to update our docs vs an external resource.

I'll let you and Cody decide to merge or close.

ifbyol: if we think this is better, I'm ok with it, but I wouldn't assume that someone has read the okteto-community guides, like using the name `okteto-letsencrypt`.

I would just say that if you already have created the certificate in the cluster in a secret, you have to use the following configuration, and use a placeholder. 

pinging @codyjlandstrom explicitly for review

codyjlandstrom:
@ekline-ai Could you please give this a review?

src/content/self-hosted/install/certificates/cert-manager.mdx
codyjlandstrom: Agree with Nacho on the placeholder. Ramiro, the argument for the section is people who already know cert-manager, and those people won't have a secret called okteto-letsencrypt. Let's use your-ssl-certificate-secret like the bring-your-own page and point to spec.secretName.

codyjlandstrom (approved):

